PR #5, opened 2022-11-22
Use snapshots for blocklist revert

dvush:
Removes state copying when applying individual tx that can potentially be reverted.


core/state_processor.go
```diff
@@ -92,16 +92,24 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
...
 	evm.Reset(txContext, statedb)
 
+	snapshot := statedb.Snapshot()
 	// Apply the transaction to the current state (included in the env).
 	result, err := ApplyMessage(evm, msg, gp)
 	if err != nil {
 		return nil, err
 	}
+	if preFinalizeHook != nil {
+		err = preFinalizeHook()
```
Ruteri, line 107: Should `preFinalizeHook` accept `result`?

dvush: I agree but I decided to not add args before we actually need them.
For 1tx bundle revert we would need results, state and maybe something else too.

PR #11, opened 2022-11-23
Removes OFAC blacklist feature.

MicahZoltu:
Software built for Ethereum should be designed with the Ethereum ethos in mind, which has a strong focus on censorship resistance.  This means not including features that make it easy for people to censor.  While we cannot stop block builders from censoring, we can at least make it difficult for them to do so by not doing all of the work of writing censorious software for them.

This PR removes the blacklisting feature, thus making it so the off-the-shelf defaults used by a builder are non-censoring.

---


MicahZoltu:
Note: I believe this includes all of the changes necessary to remove this feature, but I don't have a Go dev environment setup so I can't easily be sure.  Whoever reviews this should clone it locally and make sure it compiles without error.

MicahZoltu:
PR #56 is not a replacement for this PR at all.  #56 just removes the example blacklist, while this PR removes the entire blacklist feature.  The entire point of this PR was to remove the ability to blacklist while suing off the shelf software (force people to have to maintain a fork if they want to be bad) and removing just the example does not achieve that.

I recommend this be re-opened.

PR #16, opened 2022-12-03
Add option to submit blocks only with bundles

adrianchifor:
Adding an extra option on miner so we can skip building blocks to relays if they don't include any bundles. Mempool-only blocks with naive algos like greedy or mev-geth are not competitive enough to get bought in the majority of cases, and they can become spammy on relays if more people start running builders.

```
--miner.only_bundle_blocks     (default: false)
          Build only blocks with bundles
```

cc @avalonche @Ruteri 

---
* [x]  I have seen and agree to [`CONTRIBUTING.md`](https://github.com/flashbots/builder/blob/main/CONTRIBUTING.md)


Ruteri:
If the flag were on the miner config, you could avoid all the computation as you already know you can't build a block as soon as `txpool.MevBundles` call. I think it's worth it, as we anyway have builder-specific miner flags. what do you think?

adrianchifor:
> If the flag were on the miner config, you could avoid all the computation as you already know you can't build a block as soon as `txpool.MevBundles` call. I think it's worth it, as we anyway have builder-specific miner flags. what do you think?

Nice good point, updated to miner.

cmd/utils/flags.go
```diff
@@ -1793,6 +1798,7 @@ func setMiner(ctx *cli.Context, cfg *miner.Config) {
 		}
 		cfg.AlgoType = algoType
 	}
+	cfg.OnlyBlocksWithBundles = ctx.IsSet(MinerOnlyBlocksWithBundlesFlag.Name)
```
Ruteri, line 1801: Won't this also be the case for `--miner.only_bundle_blocks=false`?

Ruteri (changes requested):
Make sure to cover all places where `MevBundles` is called (one more fillTransactionsAlgoWorker or something).

PR #19, opened 2022-12-05
Fix empty secondary relays bug

adrianchifor:
For empty strings, `strings.Split()` always returns an array with length 1 - one empty string. This was adding in a remote relay with an empty endpoint:
```
+ log.Info("Getting validators", endpoint, r.endpoint+"/relay/v1/builder/validators")

INFO [12-05|16:48:45.362] Getting validators                       endpoint=/relay/v1/builder/validators
...
ERROR[12-05|16:48:45.362] could not get validators map from relay, retrying err="Get \"/relay/v1/builder/validators\": unsupported protocol scheme \"\""
ERROR[12-05|16:48:48.364] could not get validators map from relay  err="Get \"/relay/v1/builder/validators\": unsupported protocol scheme \"\""
ERROR[12-05|16:48:48.364] could not connect to remote relay, continuing anyway err="Get \"/relay/v1/builder/validators\": unsupported protocol scheme \"\""
```

cc @Ruteri 

---

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -33,6 +33,6 @@ var DefaultConfig = Config{
 	GenesisValidatorsRoot:         "0x0000000000000000000000000000000000000000000000000000000000000000",
 	BeaconEndpoint:                "http://127.0.0.1:5052",
 	RemoteRelayEndpoint:           "",
-	SecondaryRemoteRelayEndpoints: nil,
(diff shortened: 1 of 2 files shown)


adrianchifor:
Ah lol you just pushed while I was investigating and fixing, closing :)

PR #20, opened 2022-12-07
Fix getting validators map for local relay

avalonche:
## 📝 Summary

Previously the local relay used the onForkChoice hook to update the validator mapping. However now that the hook is unused, the builder does not update the mapping when the local relay option is enabled.

## 📚 References

---

### DIFF
--- a/builder/beacon_client.go
+++ b/builder/beacon_client.go
@@ -30,8 +30,7 @@ func (b *testBeaconClient) isValidator(pubkey PubkeyHex) bool {
 func (b *testBeaconClient) getProposerForNextSlot(requestedSlot uint64) (PubkeyHex, error) {
 	return PubkeyHex(hexutil.Encode(b.validator.Pk)), nil
 }
-
-func (b *testBeaconClient) updateValidatorsMap() error {
+func (b *testBeaconClient) Start() error {
 	return nil
 }
 
@@ -41,7 +40,6 @@ type BeaconClient struct {
 	secondsInSlot uint64
 
 	mu              sync.Mutex
-	currentEpoch    uint64
 	slotProposerMap map[uint64]PubkeyHex
 
 	closeCh chan struct{}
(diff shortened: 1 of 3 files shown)


builder/beacon_client.go
```diff
@@ -51,14 +51,36 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+	nextSlotProposer, found := b.slotProposerMap[requestedSlot]
+	if !found {
+		log.Error("inconsistent proposer mapping", "currentSlot", b.currentSlot, "slotProposerMap", b.slotProposerMap)
+		return PubkeyHex(""), errors.New("inconsistent proposer mapping")
+	}
+	return nextSlotProposer, nil
+}
+
+func (b *BeaconClient) updateValidatorsMap(nextSlot uint64) error {
+	b.mu.Lock()
```
Ruteri, line 68: I think it'd be good to avoid locking `b.mu` twice, and instead require that `updateValidatorsMap` is called with it locked.

builder/beacon_client.go
```diff
@@ -51,14 +51,36 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
 }
 
 func (b *BeaconClient) getProposerForNextSlot(requestedSlot uint64) (PubkeyHex, error) {
-	/* Only returns proposer if requestedSlot is currentSlot + 1, would be a race otherwise */
+	b.updateValidatorsMap(requestedSlot)
```
Ruteri, line 54: Calling `fetchEpochProposersMap` as blocking could result in the lock and the other functions taking too long.
The proposers' map is known an epoch in advance, so you could try fetching it every half epoch to avoid corner cases around epoch change.

builder/config.go
```diff
@@ -4,6 +4,7 @@ type Config struct {
 	Enabled               bool   `toml:",omitempty"`
 	EnableValidatorChecks bool   `toml:",omitempty"`
 	EnableLocalRelay      bool   `toml:",omitempty"`
+	SlotsInEpoch		  uint64 `toml:",omitempty"`
```
Ruteri, line 7: Lint pls

builder/beacon_client.go
```diff
@@ -65,12 +67,10 @@ func (b *BeaconClient) getProposerForNextSlot(requestedSlot uint64) (PubkeyHex,
...
 func (b *BeaconClient) updateValidatorsMap(nextSlot uint64) error {
-	b.mu.Lock()
-	defer b.mu.Unlock()
-
 	b.currentSlot = nextSlot
-	nextSlotEpoch := nextSlot / 32
-	if nextSlotEpoch != b.currentEpoch {
+	nextSlotEpoch := nextSlot / b.slotsInEpoch
+	// if slot at half epoch, fetch next epoch's proposers
+	if nextSlotEpoch != b.currentEpoch || nextSlot%b.slotsInEpoch == b.slotsInEpoch/2 {
```
Ruteri, line 73: Thinking through this, this won't work if there is no validator registered at the mid-epoch slot (won't be called for that slot)

builder/beacon_client.go
```diff
@@ -51,14 +62,56 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
 }
 
 func (b *BeaconClient) getProposerForNextSlot(requestedSlot uint64) (PubkeyHex, error) {
-	/* Only returns proposer if requestedSlot is currentSlot + 1, would be a race otherwise */
 	b.mu.Lock()
 	defer b.mu.Unlock()
 
-	if b.currentSlot+1 != requestedSlot {
-		return PubkeyHex(""), errors.New("slot out of sync")
+	if (requestedSlot / b.slotsInEpoch) > b.currentEpoch {
```
Ruteri, line 68: Hm, since `b.currentEpoch` is set elsewhere, calling this function twice with the same `requestedSlot` (which is possible) causes `currEpochProposerMap` to be cycled twice which might be incorrect

builder/beacon_client.go
```diff
@@ -51,14 +62,56 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+}
+
+func (b *BeaconClient) updateValidatorsMap() error {
+	// Start regular slot updates
+	slotsPerEpoch := b.slotsInEpoch
+	durationPerSlot := time.Duration(b.secondsInSlot) * time.Second
+	durationPerEpoch := durationPerSlot * time.Duration(slotsPerEpoch)
+	currentSlot, err := fetchCurrentSlot(b.endpoint)
+	if err != nil {
+		log.Error("could not get current slot", "err", err)
```
Ruteri, line 87: This is a bit worrying, as you don't handle the error in any way and use `currentSlot` which will be 0.
Is the setup necessary? The loop will try to do the same thing in the first iteration right?

builder/beacon_client.go
```diff
@@ -51,14 +62,56 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+	if err != nil {
+		log.Error("could not fetch validators map", "err", err)
+	}
+	b.currEpochProposerMap = slotProposerMap
+	// Every half epoch request validators map
+	timer := time.NewTicker(durationPerEpoch / 2)
+	for range timer.C {
+		currentSlot, err := fetchCurrentSlot(b.endpoint)
+		if err != nil {
+			log.Error("could not get current slot", "err", err)
```
Ruteri, line 100: Shouldn't there be a `continue` here?

builder/beacon_client.go
```diff
@@ -51,14 +62,56 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+		}
+		currentEpoch := currentSlot / b.slotsInEpoch
+		if currentEpoch > b.currentEpoch {
+			b.currentEpoch = currentEpoch
```
Ruteri, line 104: I think you want to update `b.currentEpoch` only once you have the `slotProposerMap` and do it while you keep a lock to avoid problems with concurrent access or races between the updates

builder/beacon_client.go
Ruteri: Call it twice in a row with the same requestedSlot and the asignment to currEpochProposerMap will happen twice as currentEpoch is not set in this function

builder/beacon_client.go
```diff
@@ -51,14 +62,52 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+		case <-timer.C:
+			currentSlot, err := fetchCurrentSlot(b.endpoint)
+			if err != nil {
+				log.Error("could not get current slot", "err", err)
+				continue
+			}
+			b.mu.Lock()
+			currentEpoch := currentSlot / b.slotsInEpoch
+			if currentEpoch > b.currentEpoch {
+				slotProposerMap, err := fetchEpochProposersMap(b.endpoint, b.currentEpoch+1)
```
Ruteri, line 98: Definitely shouldn't fetch with the lock held though!
Lock should happen right before you set currentEpoch and nextEpochProposerMap

builder/beacon_client.go
```diff
@@ -51,14 +62,52 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+func (b *BeaconClient) updateValidatorsMap() error {
+	// Start regular slot updates
+	slotsPerEpoch := b.slotsInEpoch
+	durationPerSlot := time.Duration(b.secondsInSlot) * time.Second
+	durationPerEpoch := durationPerSlot * time.Duration(slotsPerEpoch)
+	// Every half epoch request validators map
+	timer := time.NewTicker(durationPerEpoch / 2)
+	for {
+		select {
+		case <-timer.C:
```
Ruteri, line 89: Does the ticker tick straight away? To trigger it right away (and not wait) you could set it to zero initially.
Or factor out the update routine to a function and call it before the ticker loop for initial setup (and it's okay to wait if it fails to connect).

builder/beacon_client.go
```diff
@@ -51,14 +62,54 @@ func (b *BeaconClient) isValidator(pubkey PubkeyHex) bool {
...
+			currentSlot, err := fetchCurrentSlot(b.endpoint)
+			if err != nil {
+				log.Error("could not get current slot", "err", err)
+				continue
+			}
+
+			currentEpoch := currentSlot / b.slotsInEpoch
+			if currentEpoch > b.currentEpoch {
```
Ruteri, line 97: There's actually a race for `b.currentEpoch` between this loop and `onForkchoiceUpdate`
I wonder if just keeping a most up to date slot proposer map instead of two copies would solve most if not all of those issues.
The slot proposer map contains slots from this and the next epochs, so you have a whole epoch of runway to update it (if that fails, there are bigger issues than the builder).
Instead of keeping `nextEpochProposerMap` and `currEpochProposerMap` you could just keep  `currEpochProposerMap` up to date in this loop.

Ruteri (approved):

PR #21, opened 2022-12-07
Increase sync size to 1s to avoid hitting 429 too many requests issue

madjarevicn:
## 📝 Summary

Increased a delay from 0.5s to 1s in order to avoid hitting rate limit on flashbots api. 

## 📚 References

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN cd /go-ethereum && go run build/ci.go install -static ./cmd/geth
 FROM alpine:latest
 
 RUN apk add --no-cache ca-certificates
+
 COPY --from=builder /go-ethereum/build/bin/geth ./
 
 EXPOSE 8545 8546 30303 30303/udp


builder/builder.go
```diff
@@ -81,7 +81,7 @@ func NewBuilder(sk *bls.SecretKey, ds flashbotsextra.IDatabaseService, relay IRe
 		builderPublicKey:     pk,
 		builderSigningDomain: builderSigningDomain,
 
-		limiter:       rate.NewLimiter(rate.Every(time.Millisecond), 510),
+		limiter:       rate.NewLimiter(rate.Every(time.Millisecond), 600),
```
Ruteri, line 84: I think 510 is enough above 500 to avoid issues most of the time for relays wanting 2 blocks/s
I think the bigger problem we could try to solve is that some relays limit to 1 block/s and we'd want the rate limit to be configurable in that case

builder/builder.go
```diff
@@ -286,7 +286,7 @@ func (b *Builder) runBuildingJob(slotCtx context.Context, proposerPubkey boostTy
 	}
 
 	// resubmits block builder requests every second
-	runRetryLoop(ctx, 500*time.Millisecond, func() {
+	runRetryLoop(ctx, 1*time.Second, func() {
```
Ruteri, line 289: This should not impact the relay rate limit, as it's a loop scheduling new blocks and not a loop controlling how often they are submitted
Again, would be nice to make it either dynamic based on load or configurable - changing from one hardcoded value to another won't bring much benefit

builder/builder.go
```diff
@@ -106,6 +106,9 @@ func (b *Builder) onSealedBlock(block *types.Block, ordersClosedAt time.Time, se
 
 	value := new(boostTypes.U256Str)
 	err = value.FromBig(block.Profit)
+
+	log.Info("Block profit for block", value)
```
dvush, line 110: It has to be  `log.Info("message", "value name", value)` otherwise it will print errors.

metachris:
Fyi, the rate limit is now 450 blocks / 5 minutes, so the appropriate internal limit works be 3 block submissions within 2 seconds.

PR #24, opened 2022-12-08
Add metrics for builder

avalonche:
## 📝 Summary

Add builder metrics for benchmarking performance. To accurately compare between merging algos, we need to some refactoring to provide general interfaces the merging algorithm.

Currently the metrics only records the performance for the greedy algorithm.

TODO:
- testing
- adding additional metrics we are interested in

## 📚 References

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,12 +92,25 @@ $ geth --help
 
     --miner.extradata value       
           Block extra data set by the miner (default = client version)
+   
+   METRICS
+
+   --metrics.builder value          (default: false)
+            Enable builder metrics collection and reporting
 ```
 
 Environment variables:
 ```
 BUILDER_TX_SIGNING_KEY - private key of the builder used to sign payment transaction, must be the same as the coinbase address
 ```
+
+## Metrics
+
+To enable metrics on the builder you will need to enable metrics with the flags `--metrics --metrics.addr 127.0.0.1 --metrics.builder` which will run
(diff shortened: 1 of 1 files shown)


Ruteri:
Approach looks great. Would you mind sharing example output?

metrics/metrics.go
```diff
@@ -51,6 +57,13 @@ func init() {
 				EnabledExpensive = true
 			}
 		}
+
+		for _, enabler := range builderEnablerFlags {
+			if !EnabledBuilder && flag == enabler {
+				log.Info("Enabling builder metrics collection")
+				EnabledBuilder = true
```
Ruteri, line 64: Should probably break

miner/worker.go
```diff
@@ -1720,6 +1746,9 @@ func (w *worker) simulateBundles(env *environment, bundles []types.MevBundle, pe
 	}
 
 	log.Debug("Simulated bundles", "block", env.header.Number, "allBundles", len(bundles), "okBundles", len(simulatedBundles), "time", time.Since(start))
+	if metrics.EnabledBuilder {
+		simulatedBundlesTimer.Update(time.Since(start))
```
Ruteri, line 1750: Do we want to time the individual bundles?

metrics/metrics.go
Ruteri: Makes sesne

Ruteri (approved):

PR #27, opened 2022-12-13
Block Building Algorithm Tests

lmittmann:
## 📝 Summary

This PR adds a test suite and benchmarks for the `miner.greedyBuilder` that consists of very small tx pools and bundles and their maximum validator profit. Small tx pools and bundle sets are easy to reason about and good for testing exactly one aspect of the block builder. It is also straight forward to add new test cases to the suite.

The test cases can be inflated/scaled up to produce bigger transaction pools and bundle sets which is useful for benchmarks. In the future this framework could be extended to generate tx pools and bundle sets consisting of a combination of test cases to produce more realistic input for the builder.

---

* [X] I have seen and agree to [`CONTRIBUTING.md`](https://github.com/flashbots/builder/blob/main/CONTRIBUTING.md)

### DIFF
--- a/miner/algo_common_test.go
+++ b/miner/algo_common_test.go
@@ -186,14 +186,19 @@ func genGenesisAlloc(sign signerList, contractAddr []common.Address, contractCod
 
 func genTestSetup() (*state.StateDB, chainData, signerList) {
 	config := params.AllEthashProtocolChanges
-	db := rawdb.NewMemoryDatabase()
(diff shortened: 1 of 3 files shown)


Ruteri:
Closing in favor of the rebased and simplified https://github.com/flashbots/builder/pull/32
The code attribution will be preserved there, thanks for putting together awesome new way to do tests

PR #30, opened 2022-12-18
Work on local devnet

dvush:
## 📝 Summary

This change makes working with builder on local devnet more pleasant.

* always recommit blocks when creating them
* allow using greedy builder for local devnet

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1290,6 +1290,23 @@ func (w *worker) prepareWork(genParams *generateParams) (*environment, error) {
 	return env, nil
 }
 
+func (w *worker) fillTransactionsSelectAlgo(interrupt *int32, env *environment) (error, []types.SimulatedBundle, []types.SimulatedBundle) {
+	var (
+		blockBundles []types.SimulatedBundle
+		allBundles   []types.SimulatedBundle
+		err          error
+	)
+	switch w.flashbots.algoType {
+	case ALGO_GREEDY:
+		err, blockBundles, allBundles = w.fillTransactionsAlgoWorker(interrupt, env)
+	case ALGO_MEV_GETH:
+		err, blockBundles, allBundles = w.fillTransactions(interrupt, env)
+	default:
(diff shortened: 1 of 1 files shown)


Ruteri:
Looks good, let's rebase and merge once https://github.com/flashbots/builder/pull/20 is merged (has some other necessary fixes)

Seyvick (approved):

Seyvick (approved):

miner/worker.go
```diff
@@ -541,11 +541,12 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
 			// If sealing is running resubmit a new work cycle periodically to pull in
 			// higher priced transactions. Disable this overhead for pending blocks.
 			if w.isRunning() && (w.chainConfig.Clique == nil || w.chainConfig.Clique.Period > 0) {
+				// flashbots: disable this because there can be new bundles
 				// Short circuit if no new transaction arrives.
-				if atomic.LoadInt32(&w.newTxs) == 0 {
-					timer.Reset(recommit)
-					continue
-				}
+				//if atomic.LoadInt32(&w.newTxs) == 0 {
```
Ruteri, line 546: Is this necessary?
I'm not sure if anything depends on this in PoS.
There's been a recent change that allows "no tx" blocks to be built

dvush: I think this is a good thing to have, otherwise to commit bundles you must have txs around and that makes some type of testing more hard. Also its more correct because this code does not know about bundles and assumes that if there are no txs then there is no way to create a non-empty block.

I don't think that this matters in PoS.

1. It may matter only if people use our builder as a primary node for CL + EL setup. If we use this node as a builder this code is never called.
2. Even then I don't think this matters because this is only short-circuit, meaning that in the absence of bugs in the sealing code the same output should happen but after some steps. E.g. if its prohibited by consensus to a real empty block then it should error in another place.

Ruteri (approved):

Issue #13, opened 2022-12-01
Possible Memory Leak


cjios:
#### System information

Geth version: `1.10.23-stable`
Go version: `go1.19.3`
OS & Version: Linux
Commit hash : 9ca4d1667a57d08ce035e53ea81b1413c0352770

https://github.com/flashbots/builder/blob/6cdb430e148439c155afc176bf9984e75e4d7404/miner/algo_common.go#L74

This line is causing a memory leak for me. My system has 128gb of memory, but after running the builder for around 10-12 hours the node consumes all my available memory and crashes. Commenting out this line results in the builder only using 2-3GB of memory. The following image is the pprof heap output after running the node for around 30 minutes. As you can see the trie being copied is consuming most of the heap space. I'm not sure why the garbage collector isn't clearing the space properly?

<img width="770" alt="Screen Shot 2022-12-01 at 2 24 48 PM" src="https://user-images.githubusercontent.com/95324000/205171946-b20575a7-d84e-4ac4-8e84-a63b22c241e8.png">

Ruteri:
This is intriguing as we are running the same code for weeks without OOMs, and the profiles look clean.
Can you paste the command line arguments?
This looks like it could be an issue with not stopping the state's prefetcher.

cjios:
This is how I launch the instance `./build/bin/geth --http --http.api eth,net,engine,admin --builder --builder.remote_relay_endpoint https://boost-relay.flashbots.net --builder.secret_key {secret_key}`

Ruteri:
From the screenshot, it seems those are only using 5GBs or so. Am I misreading it?

cjios:
yes, but this is only after around 30 minutes. The memory consumptions continues to scale linearly, so after 2 or 3 hours around 30gb of memory get's consumed.

Ruteri:
Alright. I think we should take a look at how the db fetcher is handled and whether it's properly closed on all branches then.
Can you compare with running `--miner.algotype greedy`?

cjios:
<img width="786" alt="Screen Shot 2022-12-14 at 4 16 54 PM" src="https://user-images.githubusercontent.com/95324000/207743016-7e4019ec-71c2-466c-94ec-aec4669b32d4.png">

This is after running the greedy builder for ~ 2 days

cjios:
[...]

PR #35, opened 2022-12-27
Stop prefetcher in applyToBaseEnv()

dvush:
## 📝 Summary

Should fix https://github.com/flashbots/builder/issues/13

Its stops prefetcher when it can be leaked.


Ruteri (approved):
I worry this is hardly the last place it could leak, but it'll definitely help

dvush:
I am very confident that this is the only place that matter.

We create an `environment` and start prefetcher in the `prepareWork` function and call `env.discard()` (that stops prefetcher) properly. 

The only problem that we have is if we swap `env.state` pointer to the copy of the state. We do it only in one place - which is here. 

The upside of this change is that it will not stop prefetcher if we don't use this method. (e.g. if we will have algorithm that works without state copying for bundles)

PR #36, opened 2023-01-02
Bundle cancellations

Ruteri:
## 📝 Summary

Introduces bundle replacement and cancellation via `replacementUuid`.
Since the replacement is tied to a specific sender, `eth_sendBundle` gets two additional optional fields: the replacement uuid and the signingAddress of the bundle submission.
`func (b *EthAPIBackend) SendBundle(ctx context.Context, txs types.Transactions, blockNumber rpc.BlockNumber, uuid uuid.UUID, signingAddress common.Address, minTimestamp uint64, maxTimestamp uint64, revertingTxHashes []common.Hash) error`.

The DB requests are done in the background, and cancellations are resolved while non-cancelable bundles are already being simulated to avoid waiting for DB to reply.
If anything goes wrong with the cancellations, the cancelable bundles are not considered.

Note: every block is now sent to the relay, as we can no longer rely on the highest-profit rule!

## 📚 References

Cancellations docs: https://github.com/flashbots/flashbots-docs/pull/317/files

---

### DIFF
--- a/flashbotsextra/database.go
+++ b/flashbotsextra/database.go
@@ -41,10 +41,10 @@ func (NilDbService) GetLatestUuidBundles(ctx context.Context, blockNum int64) ([
 type DatabaseService struct {
 	db *sqlx.DB
 
-	insertBuiltBlockStmt      *sqlx.NamedStmt
-	insertMissingBundleStmt   *sqlx.NamedStmt
-	fetchPrioBundlesStmt      *sqlx.NamedStmt
-	fetchGetLatestUuidBundles *sqlx.NamedStmt
+	insertBuiltBlockStmt          *sqlx.NamedStmt
+	insertMissingBundleStmt       *sqlx.NamedStmt
+	fetchPrioBundlesStmt          *sqlx.NamedStmt
+	fetchGetLatestUuidBundlesStmt *sqlx.NamedStmt
 }
 
 func NewDatabaseService(postgresDSN string) (*DatabaseService, error) {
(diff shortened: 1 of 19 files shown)


flashbotsextra/database.go
```diff
@@ -32,12 +34,17 @@ func (NilDbService) GetPriorityBundles(ctx context.Context, blockNum int64, isHi
...
 type DatabaseService struct {
 	db *sqlx.DB
 
-	insertBuiltBlockStmt    *sqlx.NamedStmt
-	insertMissingBundleStmt *sqlx.NamedStmt
-	fetchPrioBundlesStmt    *sqlx.NamedStmt
+	insertBuiltBlockStmt      *sqlx.NamedStmt
+	insertMissingBundleStmt   *sqlx.NamedStmt
+	fetchPrioBundlesStmt      *sqlx.NamedStmt
+	fetchGetLatestUuidBundles *sqlx.NamedStmt
```
bhakiyakalimuthu, line 47: Nitpicking :  to make it consistent 
```suggestion
	fetchGetLatestUuidBundlesStmt *sqlx.NamedStmt
```

miner/worker.go
```diff
@@ -1404,16 +1406,27 @@ func (w *worker) getSimulatedBundles(env *environment) ([]types.SimulatedBundle,
...
 	// TODO: consider interrupt
 	simBundles, err := w.simulateBundles(env, bundles, nil) /* do not consider gas impact of mempool txs as bundles are treated as transactions wrt ordering */
 	if err != nil {
-		log.Error("Failed to simulate flashbots bundles", "err", err)
+		log.Error("Failed to simulate bundles", "err", err)
 		return nil, err
 	}
 
-	return simBundles, nil
+	ccBundles := <-ccBundlesCh
```
hsouf, line 1418: Nitpick: I wouldn't separate the simulation of bundles here, as long as I don't return simBundles & simCcBundles separately. But I would add a log to give an insight on the size of simCcBundles.

cgebe:
What is actually the reason to submit the block **every** 500ms. From experience the best block is usually built in the last second of the slot time. Wouldn't it be more efficient to start the building job in the last third or quarter to reduce load on the relays?

miner/worker.go
Ruteri: The bundles coming from the channel are not ready when we start the simulation of the other bundles, it can take up to few ms for the channel to be ready.

bhakiyakalimuthu (approved):

PR #39, opened 2023-01-04
Add simulation and algo interfaces

avalonche:
## 📝 Summary

Interfaces for merging algorithm and simulation. This allows standardisation across merging and simulation of bundles to allow metric collection across different block building steps.

## 📚 References

---


core/types/transaction.go
```diff
@@ -510,11 +540,19 @@ func NewTxWithMinerFee(tx *Transaction, baseFee *big.Int) (*TxWithMinerFee, erro
...
 	minerFee := bundle.MevGasPrice
 	return &TxWithMinerFee{
-		order:    _BundleOrder{bundle},
+		order:    SimulatedBundleOrder{Bundle: bundle},
 		minerFee: minerFee,
 	}, nil
 }
 
+func NewOrderWithTxFee(order Order, baseFee *big.Int) (*TxWithMinerFee, error) {
+	if order.TxType() == Tx {
```
Ruteri, line 549: Why not `if order.AsTx() != nil`?

core/types/transaction.go
```diff
@@ -459,28 +461,56 @@ func (s TxByNonce) Len() int           { return len(s) }
...
+	Tx OrderType = iota
+	Bundle
+)
+
+type Order interface {
 	AsTx() *Transaction
-	AsBundle() *SimulatedBundle
+	AsBundle() *MevBundle
+	AsSimulatedBundle() *SimulatedBundle
+	TxType() OrderType
```
Ruteri, line 475: I think the TxType is not needed, since it's sufficient to check the cast (it's the same complexity) unless you do switch statements (but I don't think you do)

core/types/transaction.go
```diff
@@ -459,28 +461,56 @@ func (s TxByNonce) Len() int           { return len(s) }
...
 
-type _Order interface {
+type OrderType int
+
+const (
+	Tx OrderType = iota
+	Bundle
+)
+
+type Order interface {
```
Ruteri, line 471: The order interface can't hold both the non-simulated bundle and the simulated bundles, since it's used for sorting. The way the orders are passed should change, since this interface loses semantic meaning

core/types/transaction.go
```diff
@@ -510,11 +540,19 @@ func NewTxWithMinerFee(tx *Transaction, baseFee *big.Int) (*TxWithMinerFee, erro
...
+		order:    SimulatedBundleOrder{Bundle: bundle},
 		minerFee: minerFee,
 	}, nil
 }
 
+func NewOrderWithTxFee(order Order, baseFee *big.Int) (*TxWithMinerFee, error) {
+	if order.TxType() == Tx {
+		return NewTxWithMinerFee(order.AsTx(), baseFee)
+	} else {
+		return NewBundleWithMinerFee(order.AsSimulatedBundle(), baseFee)
```
Ruteri, line 552: order being a regular bundle would just break everything, the interfaces are incompatible I think

miner/algo_default.go
```diff
@@ -0,0 +1,150 @@
...
+
+func NewGreedyBuilder(chain *core.BlockChain, chainConfig *params.ChainConfig, blacklist map[common.Address]struct{}, interrupt *int32, bundleCache *BundleCache) *GreedyBuilder {
+	return &GreedyBuilder{
+		chainData:   chainData{chainConfig, chain, blacklist},
+		interrupt:   interrupt,
+		bundleCache: bundleCache,
+	}
+}
+
+func (g *GreedyBuilder) doBuildBlock(inputEnv *environment, orders []types.Order) (*environment, []types.Order) {
```
Ruteri, line 28: A simple solution to avoid having non-simulated bundles be a part of the input orders is to pass the mempool and mevbundles separately into the doBuildBlock, since if we don't simulate up front there is no meaning to the orders

Ruteri (commented):
If we make the orders hold compatible types I it looks good

miner/algo_default.go
```diff
@@ -0,0 +1,132 @@
...
+			if err != nil {
+				log.Trace("could not apply tx", "hash", tx.Hash(), "err", err)
+				continue
+			}
+			effGapPrice, err := tx.EffectiveGasTip(envDiff.baseEnvironment.header.BaseFee)
+			if err == nil {
+				log.Trace("Included tx", "EGP", effGapPrice.String(), "gasUsed", receipt.GasUsed)
+			}
+		} else if bundle := order.Bundle(); bundle != nil {
+			err := envDiff.commitBundle(bundle, b.chainData, b.interrupt)
```
Ruteri, line 73: We kind of also want to trace the commits, not just simulations (which are just commit and then revert/discard) -> maybe it would be worth having that as a part of the separation interface?

PR #41, opened 2023-01-31
miner: fix mev gas price calc when same nonce tx not in mempool

bsh98:
## 📝 Summary

Currently, if a bundle includes a transaction that has the same sender and nonce as a transaction in the mempool (i.e. attempting to replace a mempool transaction with a bundle transaction), the gas fees from the bundle transaction will not be counted towards the bundle's value.  This PR compares transaction hashes instead of transaction nonces when checking if a bundle transaction is present in the mempool. 

### Motivation

 There are cases when a searcher may want to re-use a nonce from a mempool transaction in a bundle.  For example, a searcher may want to privately replace a pending transaction or move a PGA to a bundle auction.

### Variations

Another approach would be to only count the gas fees if the bundle transaction's effective gas price `(coinbaseDiff + gasFees) / gasUsed` is greater than N% (configured price bump) larger than the mempool transaction's gas price.  This would closely follow the replacement logic of the mempool.

https://github.com/flashbots/builder/blob/8caba1f96ff5adf5647356d231a3dcab8a74fbac/core/tx_list.go#L293-L296

This would avoid allowing "cheap" cancellations/replacements.

---

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1866,12 +1866,11 @@ func (w *worker) computeBundleGas(env *environment, bundle types.MevBundle, stat
 		}
 
 		txInPendingPool := false
+		txHash := tx.Hash()
 		if accountTxs, ok := pendingTxs[from]; ok {
 			// check if tx is in pending pool
-			txNonce := tx.Nonce()
-
 			for _, accountTx := range accountTxs {
-				if accountTx.Nonce() == txNonce {
+				if accountTx.Hash() == txHash {
 					txInPendingPool = true
 					break
 				}


bsh98:
> While I love the idea, correct me if I'm wrong in thinking that a transaction can overwrite a transaction with the same nonce, removing it from the pending list. So you'd actually still count the old, replaced transactions as not having been in the pending. This would require keeping track of the dropped transactions.

Apologies for the late reply and I appreciate the response.  

Yes, you are correct in that just comparing transaction hashes could lead to a transaction not being counted as pending (missed this in my first commit).  However, is it safe to assume that a transaction could only have been dropped and replaced if it had a lower gas price than the current pending transaction?  I believe this would allow you to count a bundle transaction as known/seen if the gas price is less than or equal to the current pending transaction.  If a transaction was dropped for another reason (e.g. pool lifetime exceeded), it wouldn't be counted as pending anyways.  In other words, if a bundle includes a transaction with the same account + nonce as a pending transaction, it is marked as known unless the effective gas price is higher (i.e. effectively replacing the pending transaction).

Ruteri:
>  However, is it safe to assume that a transaction could only have been dropped and replaced if it had a lower gas price than the current pending transaction?

Yes, but we still don't want to count that gas, even if it's lower than what we have now in the mempool. This still allows stuffing as far as I can tell.

bsh98:
> Yes, but we still don't want to count that gas, even if it's lower than what we have now in the mempool.

Apologies if I'm missing something here, but that is precisely the intent of the code.  The goal is to _not_ count the gas fees if the effective gas price is less than or equal to the transaction in the mempool.  If it's safe to assume that a transaction could have only been dropped if it had a lower gas price, then we should safely be able to prevent stuffing a dropped transaction by marking all underpriced transactions as seen.  

https://github.com/flashbots/builder/blob/660deaa241b29378ea809a636be55f47f8d7c991/miner/worker.go#L1889-L1901

We should only count the gas fees if the effective gas price is strictly greater than the transaction in the mempool.

PR #45, opened 2023-03-10
Prevent uint256 overflow during conversion

avalonche:
## 📚 References

---

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -68,11 +68,8 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if hash := types.DeriveSha(block.Transactions(), trie.NewStackTrie(nil)); hash != header.TxHash {
 		return fmt.Errorf("transaction root hash mismatch (header value %x, calculated %x)", header.TxHash, hash)
 	}
-	if v.config.IsShanghai(header.Time) {
-		if header.WithdrawalsHash == nil {
-			return fmt.Errorf("withdrawals hash is missing")
-		}
-		// Withdrawals are present after the Shanghai fork.
+	// Withdrawals are present after the Shanghai fork.
+	if header.WithdrawalsHash != nil {
 		// Withdrawals list must be present in body after Shanghai.
 		if block.Withdrawals() == nil {
 			return fmt.Errorf("missing withdrawals in block body")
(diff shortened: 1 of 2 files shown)


core/block_validator.go
```diff
@@ -77,6 +77,10 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 		if hash := types.DeriveSha(block.Withdrawals(), trie.NewStackTrie(nil)); hash != *header.WithdrawalsHash {
 			return fmt.Errorf("withdrawals root hash mismatch (header value %x, calculated %x)", *header.WithdrawalsHash, hash)
 		}
+	} else {
+		if block.Withdrawals() != nil {
```
Ruteri, line 81: If we are in Shanghai, withdrawals are required but will not be checked still

core/block_validator.go
```diff
@@ -68,14 +68,21 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
-	if header.WithdrawalsHash != nil {
-		// Withdrawals list must be present in body after Shanghai.
-		if block.Withdrawals() == nil {
-			return fmt.Errorf("missing withdrawals in block body")
-		}
-		if hash := types.DeriveSha(block.Withdrawals(), trie.NewStackTrie(nil)); hash != *header.WithdrawalsHash {
-			return fmt.Errorf("withdrawals root hash mismatch (header value %x, calculated %x)", *header.WithdrawalsHash, hash)
+	if v.config.IsShanghai(header.Time) {
+		// Withdrawals are present after the Shanghai fork.
+		if header.WithdrawalsHash != nil {
```
Ruteri, line 73: This is very confusing to read, it'd be better to have an early return here.
```
if header.WithdrawalsHash == nil {
    return error...
}
```

core/block_validator.go
```diff
@@ -68,14 +68,21 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
...
+			if block.Withdrawals() == nil {
+				return fmt.Errorf("missing withdrawals in block body")
+			}
+			if hash := types.DeriveSha(block.Withdrawals(), trie.NewStackTrie(nil)); hash != *header.WithdrawalsHash {
+				return fmt.Errorf("withdrawals root hash mismatch (header value %x, calculated %x)", *header.WithdrawalsHash, hash)
+			}
+		} 
+		return fmt.Errorf("withdrawals hash is missing")
+	} else {
+		if block.Withdrawals() != nil {
```
Ruteri, line 84: I think you should also check `WithdrawalsHash` here as they should not be set pre-shanghai

core/block_validator.go
```diff
@@ -68,15 +68,25 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if hash := types.DeriveSha(block.Transactions(), trie.NewStackTrie(nil)); hash != header.TxHash {
 		return fmt.Errorf("transaction root hash mismatch (header value %x, calculated %x)", header.TxHash, hash)
 	}
-	// Withdrawals are present after the Shanghai fork.
-	if header.WithdrawalsHash != nil {
+	if v.config.IsShanghai(header.Time) {
```
Ruteri, line 71: I looked at this again, and I'd be wary of adding this check in this specific function, as it's not only used during boost relay block validation

Ruteri (approved):

PR #51, opened 2023-03-22
feat: isolate sendBundle in additional namespace

lead4good:
* [ x ] I have seen and agree to [`CONTRIBUTING.md`](https://github.com/flashbots/builder/blob/main/CONTRIBUTING.md)

### DIFF
--- a/internal/ethapi/backend.go
+++ b/internal/ethapi/backend.go
@@ -124,6 +124,9 @@ func GetAPIs(apiBackend Backend, chain *core.BlockChain) []rpc.API {
 		}, {
 			Namespace: "eth",
 			Service:   NewBundleAPI(apiBackend, chain),
+		}, {
+			Namespace: "enclave",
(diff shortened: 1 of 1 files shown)


internal/ethapi/backend.go
```diff
@@ -124,6 +124,9 @@ func GetAPIs(apiBackend Backend, chain *core.BlockChain) []rpc.API {
 		}, {
 			Namespace: "eth",
 			Service:   NewBundleAPI(apiBackend, chain),
+		}, {
+			Namespace: "enclave",
```
metachris, line 128: the idea is to have a namespace that can be exposed through the SGX SSL setup. wonder whether it'd fit better into the patch rather than in the codebase 🤔 

```suggestion
			Namespace: "protect",
```

lead4good: I can add this another patch to the geth-sgx-gramine repo, however that would be the first builder-only patch introduced there. No strong opinion, happy to follow your decision.

PR #53, opened 2023-03-23
Add SSE subscription to builder

avalonche:
## 📝 Summary

https://github.com/ethereum/beacon-APIs/pull/305 Beacon API has introduced SSE subscriptions for payload attributes such  that block building can be triggered from this event stream instead of a custom block trigger in the flashbots prysm fork.
## 📚 References

---

### DIFF
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -119,7 +119,7 @@ func NewSimulatedBackendChain(database ethdb.Database, blockchain *core.BlockCha
 	filterBackend := &filterBackend{database, blockchain, backend}
 	backend.filterSystem = filters.NewFilterSystem(filterBackend, filters.Config{})
 	backend.events = filters.NewEventSystem(backend.filterSystem, false)
-	
+
 	header := backend.blockchain.CurrentBlock()
 	block := backend.blockchain.GetBlock(header.Hash(), header.Number.Uint64())
 
--- a/consensus/ethash/flashbots_api.go
+++ b/consensus/ethash/flashbots_api.go
@@ -10,11 +10,12 @@ type FlashbotsAPI struct {
 // GetWork returns a work package for external miner.
 //
 // The work package consists of 5 strings:
-//   result[0] - 32 bytes hex encoded current block header pow-hash
-//   result[1] - 32 bytes hex encoded seed hash used for DAG
-//   result[2] - 32 bytes hex encoded boundary condition ("target"), 2^256/difficulty
-//   result[3] - hex encoded block number
-//   result[4] - hex encoded profit generated from this block
+//
+//	result[0] - 32 bytes hex encoded current block header pow-hash
+//	result[1] - 32 bytes hex encoded seed hash used for DAG
+//	result[2] - 32 bytes hex encoded boundary condition ("target"), 2^256/difficulty
+//	result[3] - hex encoded block number
+//	result[4] - hex encoded profit generated from this block
 func (api *FlashbotsAPI) GetWork() ([5]string, error) {
 	if api.ethash.remote == nil {
 		return [5]string{}, errors.New("not supported")
--- a/consensus/ethash/sealer.go
+++ b/consensus/ethash/sealer.go
@@ -340,11 +340,12 @@ func (s *remoteSealer) loop() {
 // makeWork creates a work package for external miner.
 //
 // The work package consists of 5 strings:
-//	result[0], 32 bytes hex encoded current block header pow-hash
-//	result[1], 32 bytes hex encoded seed hash used for DAG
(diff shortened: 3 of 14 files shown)


builder/builder.go
```diff
@@ -101,6 +96,15 @@ func NewBuilder(sk *bls.SecretKey, ds flashbotsextra.IDatabaseService, relay IRe
 }
 
 func (b *Builder) Start() error {
+	// Start regular payload attributes updates
+	go func() {
+		c := make(chan types.BuilderPayloadAttributes)
+		go b.beaconClient.SubscribeToPayloadAttributesEvents(c)
+		for {
```
Ruteri, line 103: Is it time to add context handling within the builder?
It would be nice if `Stop` actually stopped this for loop

builder/service.go
```diff
@@ -120,7 +120,12 @@ func Register(stack *node.Node, backend *eth.Ethereum, cfg *Config) error {
 	copy(bellatrixForkVersion[:], bellatrixForkVersionBytes[:4])
 	proposerSigningDomain := boostTypes.ComputeDomain(boostTypes.DomainTypeBeaconProposer, bellatrixForkVersion, genesisValidatorsRoot)
 
-	beaconClient := NewBeaconClient(cfg.BeaconEndpoint, cfg.SlotsInEpoch, cfg.SecondsInSlot)
+	var beaconClient IBeaconClient
+	if cfg.BeaconEndpoint != "" {
```
Ruteri, line 124: Does it make sense to run the builder service without a beacon configured?

builder/beacon_client.go
```diff
@@ -256,3 +286,59 @@ func fetchBeacon(url string, dst any) error {
 	log.Info("fetched", "url", url, "res", dst)
 	return nil
 }
+
+// SubscribeToPayloadAttributesEvents subscribes to payload attributes events to validate fields such as prevrandao and withdrawals
+func (b *BeaconClient) SubscribeToPayloadAttributesEvents(payloadAttrC chan types.BuilderPayloadAttributes) {
+	payloadAttributesResp := &struct {
```
Ruteri, line 292: This got to be put somewhere as a regular structure :)

builder/beacon_client.go
```diff
@@ -256,3 +286,59 @@ func fetchBeacon(url string, dst any) error {
...
+			} `json:"payload_attributes"`
+		} `json:"data"`
+	}{}
+
+	eventsURL := fmt.Sprintf("%s/eth/v1/events?topics=payload_attributes", b.endpoint)
+	log.Info("subscribing to payload_attributes events")
+
+	for {
+		client := sse.NewClient(eventsURL)
+		err := client.SubscribeRaw(func(msg *sse.Event) {
```
Ruteri, line 311: Would it be better if the callback was a regular function?

builder/builder.go
```diff
@@ -101,10 +98,25 @@ func NewBuilder(sk *bls.SecretKey, ds flashbotsextra.IDatabaseService, relay IRe
...
+			select {
+			case <-b.stop:
+				break
+			default:
+				payloadAttributes := <-c
```
Ruteri, line 110: this should be a concurrent select
```
			select {
			case <-b.stop:
				break
			case payloadAttributes <-c:
                                b.OnPayloadAttribute(&payloadAttributes)
```

otherwise it can get indefinitely stuck waiting for payload attributes

builder/service.go
Ruteri: Why though?
The old code seems to always construct a beacon client

builder/beacon_client.go
```diff
@@ -256,3 +286,59 @@ func fetchBeacon(url string, dst any) error {
...
+			err := json.Unmarshal(msg.Data, &payloadAttributesResp)
```
Ruteri, line 312: Is this a double reference - && struct?
I'd much prefer if we had the structure defined as a standalone for ease of reasoning

builder/builder.go
```diff
@@ -277,18 +302,19 @@ func (b *Builder) OnPayloadAttribute(attrs *types.BuilderPayloadAttributes) erro
 	b.slotMu.Lock()
 	defer b.slotMu.Unlock()
 
-	if b.slot != attrs.Slot {
-		if b.slotCtxCancel != nil {
-			b.slotCtxCancel()
-		}
+	// Forcibly cancel previous building job, build on top of reorgable blocks as this is the behaviour relays expect.
```
dvush, line 305: Can you please explain why "buidling on multiple tips" is no longer necessary?

Ruteri: It is, it just doesn't work with how relays are set up. This will come back shortly after Capella.

Ruteri (approved):

PR #69, opened 2023-05-09
Ssz encoded request

avalonche:
## 📝 Summary

Submit ssz encoded block submissions if ssz is enabled on the relay specified in the config. defaults to json if relay is not listed in config.

## 📚 References

---


builder/service.go
```diff
@@ -144,9 +155,27 @@ func Register(stack *node.Node, backend *eth.Ethereum, cfg *Config) error {
 		localRelay = NewLocalRelay(relaySk, beaconClient, builderSigningDomain, proposerSigningDomain, ForkData{cfg.GenesisForkVersion, cfg.BellatrixForkVersion, cfg.GenesisValidatorsRoot}, cfg.EnableValidatorChecks)
 	}
 
+	relayConfigs := make(map[string]RelayConfig)
+	if cfg.RelayConfigFile != "" {
+		bytes, err := os.ReadFile(cfg.RelayConfigFile)
+		if err != nil {
+			log.Info("Failed to read relay config file, will use default settings", "err", err)
```
metachris, line 162: should the builder not fail here? i.e. if a config file was explicitly requested, it should never fall back to other configuration, as it's not what the user specifically requested 🤔

builder/service.go
```diff
@@ -144,9 +155,27 @@ func Register(stack *node.Node, backend *eth.Ethereum, cfg *Config) error {
...
+		} else {
+			var configs []RelayConfig
+			if err := json.Unmarshal(bytes, &configs); err != nil {
+				log.Info("Failed to unmarshal relay config, will use default settings", "err", err)
```
metachris, line 166: same here, if config fails, don't use a non-specified config?

builder/config.go
```diff
@@ -41,4 +42,10 @@ var DefaultConfig = Config{
 	RemoteRelayEndpoint:           "",
 	SecondaryRemoteRelayEndpoints: nil,
 	ValidationBlocklist:           "",
+	RelayConfigFile:               "",
+}
+
+type RelayConfig struct {
```
Wazzymandias, line 48: can we add small concise godoc for this struct?

builder/local_relay.go
```diff
@@ -105,6 +105,10 @@ func (r *LocalRelay) SubmitBlockCapella(msg *capella.SubmitBlockRequest, _ Valid
 	return r.submitBlockCapella(msg)
 }
 
+func (r *LocalRelay) Config() RelayConfig {
```
Wazzymandias, line 108: could we add a concise comment here? specifically i'd like us to clarify why it's okay for LocalRelay to return empty RelayConfig

builder/utils.go
```diff
@@ -0,0 +1,34 @@
+package builder
```
Wazzymandias, line 1: any reason why this is called `utils.go` vs `ssz.go` ?

builder/relay.go
```diff
@@ -134,7 +139,11 @@ func (r *RemoteRelay) Stop() {}
 
 func (r *RemoteRelay) SubmitBlock(msg *boostTypes.BuilderSubmitBlockRequest, _ ValidatorData) error {
```
dvush, line 140: just an observation: submit block should have ctx here. The same ctx that is used for the block building job

no need to fix in thir pr

builder/relay.go
```diff
@@ -151,12 +160,35 @@ func (r *RemoteRelay) SubmitBlock(msg *boostTypes.BuilderSubmitBlockRequest, _ V
...
 	log.Info("submitting block to remote relay", "endpoint", r.endpoint)
-	code, err := server.SendHTTPRequest(context.TODO(), *http.DefaultClient, http.MethodPost, r.endpoint+"/relay/v1/builder/blocks", msg, nil)
-	if err != nil {
-		return fmt.Errorf("error sending http request to relay %s. err: %w", r.endpoint, err)
+
+	endpoint := r.endpoint + "/relay/v1/builder/blocks"
+	if r.cancellationsEnabled {
+		endpoint = endpoint + "?cancellations=true"
 	}
-	if code > 299 {
```
dvush, line 158: This is not your change but I wonder why we use 299 instead of `!= http.StatusOk`

builder/config.go
```diff
@@ -19,6 +19,8 @@ type Config struct {
 	RemoteRelayEndpoint           string   `toml:",omitempty"`
 	SecondaryRemoteRelayEndpoints []string `toml:",omitempty"`
 	ValidationBlocklist           string   `toml:",omitempty"`
+	RelayConfigFile               string   `toml:",omitempty"`
```
dvush, line 22: I understand the appeal of adding relay config to a separate file, but I feel that this complicates config setup.

Right now, we have two ways of setting up builder config:
1. toml
2. cli params
you have to look at two places to know the builder config. (one exception is blacklist.json, but this one more like "data" thing than basic config)

Relay config file will be part of config separated and moved to json.

Why don't we map relay config into config way that we already have?

1. in toml relay config will be in place
2. in cli relay config can be added as an optional modification to the relay url that will look like:
`--builder.remote_relay_endpoint='URL;ssz;gzip'`
and the same thing with secondary endpoint, but they will be comma separated like today.


cc @metachris @Wazzymandias

builder/relay.go
metachris: had that for a long time, because any 2xx error is okay, and starting with 300 is an error

builder/config.go
metachris: i think that would make sense, and it's probably better than using multiple config files

metachris:
> what do you mean by previous configuration? if started with a previous config the node will need to be restarted to load the new config

just was wondering if this is backwards compatible to any previous configuration (i.e. with the args that worked previously)

builder/utils.go
Wazzymandias: that makes sense - I don't feel too strongly about this but I think it would be nice if we determine another name for such a file. `utils` is an anti pattern in Go. Many projects still use it but we should try to stick to idioms when they make sense.

https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

builder/relay.go
Wazzymandias: `because any 2xx error is okay`

I don't know if this is true. There are certain 2XX responses that yield partial or no content. I don't think it is good practice to handle such events as if they were valid even if the request itself was "successful" 

Such conditional also does not check for 1XX responses. It would be better in our application layer to handle these scenarios with more granularity otherwise it can be very difficult to debug and yield misleading results

builder/local_relay.go
Wazzymandias: > way to not specify a config for all IRelay as only RemoteRelay is using the config

We could probably add a method that exposes type of relay that we use to determine whether it has config or not.

That being said, I'm wondering why we can't put information into the config like `enableBeaconChecks`, `ForkData`, `*bls.SecretKey`, etc into the config so that it provides some value even for local relay

builder/relay.go
metachris: makes sense. let's revisit this another time though as it's outside the scope of this PR

metachris:
As discussed we should ship with cancellations disabled by default

dvush (approved):

builder/local_relay.go
Wazzymandias: yeah exactly

PR #70, opened 2023-05-09
[build-295] Improve Builder submission rate

Wazzymandias:
## 📝 Summary

- Fix rate limit bug that allowed more builder submissions than desired
- Expose `FLASHBOTS_BUILDER_RATE_LIMIT_DURATION` and `FLASHBOTS_BUILDER_BURST_LIMIT` environment variables
- Update builder submission logic such that submissions are sent near end of slot 

## 🧪 Testing
- Testing was done with test builder deployed on production mainnet 

### Without builder submission change
image

### With builder submission change
image

### TL;DR
- Prior to change, builder submissions happened roughly between ~t-11 to ~t+3 where t is the time of slot s
- After change, builder submissions happen roughly between ~t-4 to ~t+3, which is desired

## 📚 References

Untitled

---


builder/resubmit_utils.go
```diff
@@ -9,13 +9,19 @@ import (
...
-func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func()) {
+func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func(), submitTime *time.Time) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-updateSignal:
-			res := limiter.Reserve()
+			var res *rate.Reservation
+			if submitTime != nil {
```
metachris, line 19: it'd be slighly easier to comprehend if the first comparison is `==` (i.e. `if submitTime == nil` ...)

builder/builder.go
```diff
@@ -29,6 +29,10 @@ import (
 	boostTypes "github.com/flashbots/go-boost-utils/types"
 )
 
+const (
+	RateLimitInterval = 500 * time.Millisecond
```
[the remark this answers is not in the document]
Wazzymandias: yeah, that's what we're doing: [here](https://github.com/flashbots/builder/pull/70/files#diff-7226de6d186cf4d92743da0f3214f84e76afe158f247cbb43ff2808a08e59b04R205)

builder/service.go
```diff
@@ -173,6 +177,34 @@ func Register(stack *node.Node, backend *eth.Ethereum, cfg *Config) error {
...
 
+	var (
+		limiter *rate.Limiter
+		burst   = 10
+		// FLASHBOTS_BUILDER_RATE_LIMIT_DURATION will accept valid time.Duration values.
+		// A duration string is a possibly signed sequence of
+		// decimal numbers, each with optional fraction and a unit suffix,
+		// such as "300ms", "-1.5h" or "2h45m".
+		// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+		builderRateLimit  = os.Getenv("FLASHBOTS_BUILDER_RATE_LIMIT_DURATION")
```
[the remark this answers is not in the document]
Wazzymandias: sure - is there anything i should elaborate on [from what's in the README?](https://github.com/flashbots/builder/pull/70/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R118)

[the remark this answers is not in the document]
metachris: yeah, i think maybe elaborate on what are "burst events"

builder/builder.go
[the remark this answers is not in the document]
Wazzymandias: > but I assume this is just the default that'll be used if no env var given?

yes exactly

> i think the link is not working
apologies, I was referring to line 188 in `builder/service.go` where we accept an environment variable that's used to instantiate the rate limiter. If the environment variable is not provided, we would use this default

I can update the name, that makes sense 👍🏾

builder/resubmit_utils.go
Wazzymandias: i agree, updated

builder/builder.go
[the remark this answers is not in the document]
Wazzymandias: updated

builder/builder.go
```diff
@@ -73,11 +77,17 @@ type Builder struct {
...
-func NewBuilder(sk *bls.SecretKey, ds flashbotsextra.IDatabaseService, relay IRelay, builderSigningDomain boostTypes.Domain, eth IEthereumService, dryRun bool, ignoreLatePayloadAttributes bool, validator *blockvalidation.BlockValidationAPI, beaconClient IBeaconClient) *Builder {
+func NewBuilder(sk *bls.SecretKey, ds flashbotsextra.IDatabaseService, relay IRelay, builderSigningDomain boostTypes.Domain,
+	eth IEthereumService, dryRun bool, ignoreLatePayloadAttributes bool, validator *blockvalidation.BlockValidationAPI,
+	beaconClient IBeaconClient, limiter *rate.Limiter) *Builder {
 	pkBytes := bls.PublicKeyFromSecretKey(sk).Compress()
 	pk := boostTypes.PublicKey{}
 	pk.FromSlice(pkBytes)
 
+	if limiter == nil {
+		limiter = rate.NewLimiter(rate.Every(RateLimitIntervalDefault), 10)
```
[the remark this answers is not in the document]
Wazzymandias: agreed, updated

README.md
```diff
@@ -115,6 +115,9 @@ $ geth --help
 Environment variables:
 ```
 BUILDER_TX_SIGNING_KEY - private key of the builder used to sign payment transaction, must be the same as the coinbase address
+FLASHBOTS_BUILDER_RATE_LIMIT_DURATION - determines rate limit of events processed by builder; a duration string is a
```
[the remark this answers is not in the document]
dvush: also, every config should be available in toml too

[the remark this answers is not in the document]
Wazzymandias: makes sense, updated. Please feel free to un-resolve if additional changes are necessary

builder/service.go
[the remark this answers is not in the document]
Wazzymandias: this changed a bit based on discussion found [here](https://github.com/flashbots/builder/pull/70#discussion_r1189577008)

I've updated the documentation to hopefully be more clear. Please feel free to un-resolve if there are still concerns

builder/resubmit_utils.go
```diff
@@ -9,13 +9,28 @@ import (
...
+func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func(), submitTime *time.Time) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-updateSignal:
-			res := limiter.Reserve()
+			var res *rate.Reservation
+			if submitTime == nil {
+				res = limiter.Reserve()
```
dvush, line 20: since you reserve in the end of both branches of the if-else you can move it out and put it after

builder/resubmit_utils.go
```diff
@@ -9,13 +9,28 @@ import (
...
+			} else {
+				now := time.Now()
+				sleepTime := submitTime.UTC().UnixMilli() - now.UTC().UnixMilli()
+				log.Debug("resubmit loop",
```
dvush, line 24: nit: it look more like `log.Trace` to me

cmd/utils/flags.go
```diff
@@ -1618,6 +1645,8 @@ func SetBuilderConfig(ctx *cli.Context, cfg *builder.Config) {
 	cfg.RemoteRelayEndpoint = ctx.String(BuilderRemoteRelayEndpoint.Name)
 	cfg.SecondaryRemoteRelayEndpoints = strings.Split(ctx.String(BuilderSecondaryRemoteRelayEndpoints.Name), ",")
 	cfg.ValidationBlocklist = ctx.String(BuilderBlockValidationBlacklistSourceFilePath.Name)
+	cfg.BuilderRateLimitDuration = ctx.String(BuilderRateLimitDuration.Name)
+	cfg.BuilderRateLimitMaxBurst = ctx.Int(BuilderRateLimitMaxBurst.Name)
```
dvush, line 1649: what about 500ms that is used in 
`runRetryLoop(ctx, 500*time.Millisecond, func() {`
I remember we were talking about that too and this one might actually make a difference in builder performance

builder/resubmit_utils.go
```diff
@@ -9,13 +9,28 @@ import (
...
 		case <-ctx.Done():
 			return
 		case <-updateSignal:
-			res := limiter.Reserve()
+			var res *rate.Reservation
+			if submitTime == nil {
+				res = limiter.Reserve()
+			} else {
+				now := time.Now()
+				sleepTime := submitTime.UTC().UnixMilli() - now.UTC().UnixMilli()
```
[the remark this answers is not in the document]
Wazzymandias: > why this instead of just time.Sub without converting to millis and back (in sleep)

I was debugging timestamp issue because input `attrs.Timestamp` was getting set to the year `1970` so `time.Sub` was incorrect. But that's fixed now so I'll update 👍🏾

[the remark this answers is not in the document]
Wazzymandias: updated

builder/resubmit_utils.go
Wazzymandias: good point, updated

cmd/geth/main.go
```diff
@@ -175,6 +175,9 @@ var (
 		utils.BuilderBeaconEndpoints,
 		utils.BuilderRemoteRelayEndpoint,
 		utils.BuilderSecondaryRemoteRelayEndpoints,
+		utils.BuilderRateLimitDuration,
```
Ruteri, line 178: Would be nice to also update the readme flags, usually best to just run `geth --help` and copy the relevant new flags

builder/builder.go
```diff
@@ -73,22 +82,33 @@ type Builder struct {
...
 func NewBuilder(args BuilderArgs) *Builder {
 	pkBytes := bls.PublicKeyFromSecretKey(args.sk).Compress()
 	pk := boostTypes.PublicKey{}
 	pk.FromSlice(pkBytes)
 
+	if args.limiter == nil {
+		args.limiter = rate.NewLimiter(rate.Every(RateLimitIntervalDefault), RateLimitBurstDefault)
+	}
+
+	if args.builderBlockResubmitInterval == 0 {
```
Ruteri, line 108: I wonder if this shouldn't be the caller's responsibility (same for the above)

builder/builder.go
```diff
@@ -129,11 +150,21 @@ func (b *Builder) Start() error {
 				} else if payloadAttributes.Slot == currentSlot {
 					// Subsequent sse events should only be canonical!
 					if !b.ignoreLatePayloadAttributes {
-						b.OnPayloadAttribute(&payloadAttributes)
+						err := b.OnPayloadAttribute(&payloadAttributes)
```
Ruteri, line 153: The function already handles logging, I don't think this is necessary.
Otherwise you should also remove the logging from the hook, and wrap the errors instead.

builder/builder.go
```diff
@@ -158,7 +189,7 @@ func (b *Builder) onSealedBlock(block *types.Block, blockValue *big.Int, ordersC
 		}
 	}
 
-	log.Info("submitted block", "slot", attrs.Slot, "value", blockValue.String(), "parent", block.ParentHash, "hash", block.Hash(), "#commitedBundles", len(commitedBundles))
+	log.Info("submitted block", "slot", attrs.Slot, "value", blockValue.String(), "parent", block.ParentHash, "hash", block.Hash(), "#commitedBundles", len(commitedBundles), "blockTimestamp", uint64(attrs.Timestamp))
```
Ruteri, line 192: Is timestamp necessary if we already have the slot?
Same in a couple different places

builder/builder.go
```diff
@@ -378,11 +411,18 @@ func (b *Builder) runBuildingJob(slotCtx context.Context, proposerPubkey boostTy
...
+	// Avoid submitting early into a given slot. For example if slots have 12 second interval, submissions should
+	// not begin until 8 seconds into the slot.
+	slotTime := time.Unix(int64(attrs.Timestamp), 0).UTC()
+	submitTime := slotTime.Add(-SubmissionDelaySecondsDefault)
+
+	ctx = context.WithValue(ctx, key("slot"), attrs.Slot)
+	ctx = context.WithValue(ctx, key("timestamp"), uint64(attrs.Timestamp))
 	// Empties queue, submits the best block for current job with rate limit (global for all jobs)
-	go runResubmitLoop(ctx, b.limiter, queueSignal, submitBestBlock)
+	go runResubmitLoop(ctx, b.limiter, queueSignal, submitBestBlock, &submitTime)
```
Ruteri, line 422: We pass in some of the values as arguments, and some as context values - any reason not to pass in everything as normal values?

builder/resubmit_utils.go
```diff
@@ -9,13 +9,28 @@ import (
...
-func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func()) {
+func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func(), submitTime *time.Time) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-updateSignal:
-			res := limiter.Reserve()
+			var res *rate.Reservation
+			if submitTime == nil {
```
Ruteri, line 19: Can `submitTime` be nil? If not, maybe it would be better to assert it's not nil and require that it be passed in

builder/resubmit_utils.go
```diff
@@ -9,13 +9,28 @@ import (
...
+			} else {
+				now := time.Now()
+				sleepTime := submitTime.UTC().UnixMilli() - now.UTC().UnixMilli()
+				log.Debug("resubmit loop",
+					"now-sec", now.UTC().Unix(),
+					"slot", ctx.Value(key("slot")),
+					"slot-sec", submitTime.Add(4*time.Second).Unix(),
+					"delta-now-from-slot-ms", submitTime.Add(4*time.Second).Sub(now).Milliseconds(),
+					"block-time-sec", ctx.Value(key("timestamp")))
+				time.Sleep(time.Duration(sleepTime) * time.Millisecond)
```
Ruteri, line 30: Small nit: sleep could be interrupted by context cancellation, maybe it makes sense to select on `time.After` and `context.Done()`

builder/resubmit_utils_test.go
```diff
@@ -43,7 +43,7 @@ func TestResubmitUtils(t *testing.T) {
 			subAll = append(subAll, submission{time.Now(), subBest})
 			subLast = subBest
 		}
-	})
+	}, nil)
```
Ruteri, line 46: Could also be equivalent to `time.Now()`

builder/resubmit_utils.go
Ruteri, line 23: `sleepTime` could be negative here

cmd/utils/flags.go
```diff
@@ -804,12 +804,50 @@ var (
 		Value:    "",
 		Category: flags.BuilderCategory,
 	}
+
+	// Builder rate limit settings
+
+	// BuilderRateLimitDuration determines rate limit of events processed by builder.
```
Ruteri, line 810: Shouldn't this comment be omitted since it's already in the `Usage`?

cmd/utils/flags.go
```diff
@@ -804,12 +804,50 @@ var (
...
+	// Input is the time interval between events, which is converted to a rate.Limit
+	// which represents the number of events per second.
+	BuilderRateLimitDuration = &cli.StringFlag{
```
Ruteri, line 813: I find defaults helpful, since they showcase the format

cmd/utils/flags.go
```diff
@@ -804,12 +804,50 @@ var (
...
+	// Additional details can be found on rate.Limiter documentation: https://pkg.go.dev/golang.org/x/time/rate#Limiter
+	BuilderRateLimitMaxBurst = &cli.IntFlag{
+		Name:     "builder.rate_limit_max_burst",
+		Usage:    "Determines the maximum number of burst events the builder can accommodate at any given point in time.",
+		EnvVars:  []string{"FLASHBOTS_BUILDER_RATE_LIMIT_MAX_BURST"},
+		Value:    builder.RateLimitBurstDefault,
+		Category: flags.BuilderCategory,
+	}
+
+	// BuilderBlockResubmitInterval determines the interval at which builder will resubmit block submissions.
```
Ruteri, line 834: Comment is already in `Usage`

cmd/utils/flags.go
```diff
@@ -804,12 +804,50 @@ var (
...
+	}
+
 	BuilderEnableCancellations = &cli.BoolFlag{
 		Name:     "builder.cancellations",
 		Usage:    "Enable cancellations for the builder",
 		Category: flags.BuilderCategory,
 	}
+
 	// RPC settings
+
```
Ruteri, line 850: Since those are not our flags, I'd probably suggest not to add this new line as a good measure to avoid rebase conflicts as much as possible

cmd/utils/flags.go
```diff
@@ -2391,7 +2431,7 @@ func DialRPCWithHeaders(endpoint string, headers []string) (*rpc.Client, error)
 	}
 	var opts []rpc.ClientOption
 	if len(headers) > 0 {
-		var customHeaders = make(http.Header)
+		customHeaders := make(http.Header)
```
Ruteri, line 2434: Hm, why? I don't think we want to change upstream code unless necessary

builder/resubmit_utils.go
Ruteri, line 30: I'd add a comment explaining why we can indeed sleep here, ie who submits to the channel and that it's non-blocking

builder/builder.go
Wazzymandias: > I wonder if this shouldn't be the caller's responsibility (same for the above)

I have it in both places. In general I agree with you re: caller's responsibility. However I would want to return an error for invalid inputs but the function signature would need to be updated. I didn't want to do that work in this PR - happy to discuss

builder/builder.go
Wazzymandias: > Is timestamp necessary if we already have the slot?

Good question - out of an abundance of caution I added it separate to sanity check that the timestamp indeed match the slot. I don't know if we're checking elsewhere that timestamp always matches slot. This was mostly useful for me debugging so open to removing it

builder/builder.go
Wazzymandias: good catch, discussed offline - wrapped the errors and removed logging from the hook, please feel free to un-resolve if i missed anything

builder/builder.go
Wazzymandias: I've removed the context values because that was meant for investigating a separate issue unrelated to this PR. To reduce scope I'll add those changes to a separate branch, so resolving this

cmd/utils/flags.go
Wazzymandias: removed

cmd/utils/flags.go
Wazzymandias: fair point, updated

cmd/utils/flags.go
Wazzymandias: good catch, didn't think about it affecting upstream code, reverted

builder/builder.go
Wazzymandias: i've removed timestamp since it was more for analysis of a separate issue unrelated to this PR - so going to have it on separate branch, removed it here

builder/resubmit_utils.go
Wazzymandias: i've updated the logic here based on other feedback, please feel free to unresolve if there are still concerns

builder/resubmit_utils.go
Wazzymandias: good point, based on your other comment i've refactored this, please feel free to unresolve if there are still concerns

builder/resubmit_utils.go
Wazzymandias: good catch, updated

builder/resubmit_utils_test.go
Wazzymandias: updated

builder/resubmit_utils.go
Wazzymandias: updated

cmd/utils/flags.go
Wazzymandias: good point, updated

builder/resubmit_utils.go
```diff
@@ -9,13 +9,45 @@ import (
...
+	if submitTime.IsZero() {
+		log.Warn("skipping resubmit loop - zero submit time found")
+		return
+	}
+
+	var (
+		waitUntilSubmitTime = func(waitUntil time.Time) (ok bool) {
+			now := time.Now().UTC()
+			if waitUntil.UTC().Before(now) {
+				waitUntil = now
```
Ruteri, line 22: shouldn't this just be `return true`?

cmd/utils/flags.go
Ruteri: I see this comment is still present both as a comment and in the flag description

dvush (approved):

cmd/geth/main.go
Wazzymandias: ohh i thought you meant just the env var, my bad, updated!

builder/builder.go
Wazzymandias: > I'd make it the responsibility of the caller still, since you don't really validate the value

For now what's wrong with having it in both places? I'm worried that without the defensive check in the receiver, we can easily have a resubmit interval of 0. My preference would rather be to refactor the function signature in the future such that a 0 value returns an error.

builder/resubmit_utils.go
Wazzymandias: not necessarily i think - in the time the function is invoked i want to make sure the context hasn't been cancelled

builder/resubmit_utils.go
```diff
@@ -9,13 +9,45 @@ import (
...
 // runResubmitLoop checks for update signal and calls submit respecting provided rate limiter and context
-func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal chan struct{}, submit func()) {
+func runResubmitLoop(ctx context.Context, limiter *rate.Limiter, updateSignal <-chan struct{}, submit func(), submitTime time.Time) {
+	if submitTime.IsZero() {
+		log.Warn("skipping resubmit loop - zero submit time found")
+		return
+	}
+
+	var (
+		waitUntilSubmitTime = func(waitUntil time.Time) (ok bool) {
```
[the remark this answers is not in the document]
Wazzymandias: keeping it within func since it's not used elsewhere

Agreed on returning context, updated signature to return bool and context error

builder/resubmit_utils.go
```diff
@@ -9,13 +9,45 @@ import (
...
+			case <-time.After(sleepTime):
+				ok = true
+			}
+			return ok
+		}
+
+		canContinue bool
+	)
+
+	if canContinue = waitUntilSubmitTime(submitTime); !canContinue {
```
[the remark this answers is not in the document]
Wazzymandias: agreed, updated

avalonche (approved):

PR #72, opened 2023-06-01
Update go-boost-utils and type upgrades

avalonche:
## 📝 Summary

Upgrading go-boost-utils so we can move away entirely from go-boost-util types and use attestant types instead

## 📚 References

---


Wazzymandias (approved):
lgtm! I think we should deploy before we merge so conditionally approved 🚢

Ruteri (approved):

PR #76, opened 2023-06-14
Add support for additional block building algorithm

Wazzymandias:
## 📝 Summary

- This PR adds support for a new block building algorithm type, called `greedy-buckets` 
- The new algorithm attempts to insert transactions by both effective gas price and profit. This is in contrast to the `greedy` algorithm which simply uses effective gas price 

## Testing
- Tested locally and on production mainnet
- Observed a healthy number of higher bids with the new algorithm, screenshot below compares how often the new algorithm in blue outbid the existing one in orange: 
image

- Below is a chart that compares top bid of each algorithm per block building attempt
image

## 📚 References

- [The Knapsack Problem with Conflict Graphs](https://jgaa.info/accepted/2009/PferschySchauer2009.13.2.pdf)

---

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -713,18 +713,12 @@ func (t *TransactionsByPriceAndNonce) ShiftAndPushByAccountForTx(tx *Transaction
 	}
 
 	acc, _ := Sender(t.signer, tx)
-	txs, ok := t.txs[acc]
-	if !ok || len(txs) < 1 {
-		return
+	if txs, exists := t.txs[acc]; exists && len(txs) > 0 {
+		if wrapped, err := NewTxWithMinerFee(txs[0], t.baseFee); err == nil {
+			t.txs[acc] = txs[1:]
+			heap.Push(&t.heads, wrapped)
+		}
 	}
-
-	wrapped, err := NewTxWithMinerFee(txs[0], t.baseFee)
-	if err != nil || wrapped == nil {
-		return
-	}
-
-	t.txs[acc] = txs[1:]
-	heap.Push(&t.heads, wrapped)
 }
 
 func (t *TransactionsByPriceAndNonce) Push(tx *TxWithMinerFee) {
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -543,6 +543,22 @@ func (t *TxWithMinerFee) Profit(baseFee *big.Int, gasUsed uint64) *big.Int {
 	}
 }
 
+// SetPrice sets the miner fee of the wrapped transaction.
+func (t *TxWithMinerFee) SetPrice(price *big.Int) {
+	t.minerFee.Set(price)
+}
+
+// SetProfit sets the profit of the wrapped transaction.
+func (t *TxWithMinerFee) SetProfit(profit *big.Int) {
+	if bundle := t.Bundle(); bundle != nil {
+		bundle.TotalEth.Set(profit)
+	} else if sbundle := t.SBundle(); sbundle != nil {
+		sbundle.Profit.Set(profit)
+	} else {
+		panic("SetProfit called on unsupported order type")
+	}
+}
+
 // NewTxWithMinerFee creates a wrapped transaction, calculating the effective
 // miner gasTipCap if a base fee is provided.
 // Returns error in case of a negative effective miner gasTipCap.
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -525,8 +525,12 @@ func (t *TxWithMinerFee) Price() *big.Int {
 	return new(big.Int).Set(t.minerFee)
 }
 
-func (t *TxWithMinerFee) Profit() *big.Int {
-	if bundle := t.Bundle(); bundle != nil {
+func (t *TxWithMinerFee) Profit(baseFee *big.Int) *big.Int {
+	if tx := t.Tx(); tx != nil {
+		profit := new(big.Int).Sub(tx.GasPrice(), baseFee)
+		profit.Mul(profit, new(big.Int).SetUint64(tx.Gas()))
+		return profit
+	} else if bundle := t.Bundle(); bundle != nil {
 		return bundle.TotalEth
 	} else if sbundle := t.SBundle(); sbundle != nil {
 		return sbundle.Profit
@@ -721,6 +725,14 @@ func (t *TransactionsByPriceAndNonce) ShiftAndPushByAccountForTx(tx *Transaction
 
 	t.txs[acc] = txs[1:]
 	heap.Push(&t.heads, wrapped)
+}
+
+func (t *TransactionsByPriceAndNonce) Push(tx *TxWithMinerFee) {
+	if tx == nil {
+		return
+	}
+
+	heap.Push(&t.heads, tx)
 }
 
 // Pop removes the best transaction, *not* replacing it with the next one from
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -683,6 +683,26 @@ func (t *TransactionsByPriceAndNonce) Shift() {
 	heap.Pop(&t.heads)
 }
 
+func (t *TransactionsByPriceAndNonce) ShiftAndPushByAccountForTx(tx *Transaction) {
+	if tx == nil {
+		return
+	}
+
+	acc, _ := Sender(t.signer, tx)
+	txs, ok := t.txs[acc]
+	if !ok || len(txs) < 1 {
+		return
+	}
+
+	wrapped, err := NewTxWithMinerFee(txs[0], t.baseFee)
+	if err != nil || wrapped == nil {
+		return
+	}
+
+	t.txs[acc] = txs[1:]
+	heap.Push(&t.heads, wrapped)
+}
+
 // Pop removes the best transaction, *not* replacing it with the next one from
 // the same account. This should be used when a transaction cannot be executed
 // and hence all subsequent ones should be discarded from the same account.
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -525,10 +525,14 @@ func (t *TxWithMinerFee) Price() *big.Int {
 	return new(big.Int).Set(t.minerFee)
 }
 
-func (t *TxWithMinerFee) Profit(baseFee *big.Int) *big.Int {
+func (t *TxWithMinerFee) Profit(baseFee *big.Int, gasUsed uint64) *big.Int {
 	if tx := t.Tx(); tx != nil {
 		profit := new(big.Int).Sub(tx.GasPrice(), baseFee)
-		profit.Mul(profit, new(big.Int).SetUint64(tx.Gas()))
+		if gasUsed != 0 {
+			profit.Mul(profit, new(big.Int).SetUint64(gasUsed))
+		} else {
+			profit.Mul(profit, new(big.Int).SetUint64(tx.Gas()))
+		}
 		return profit
 	} else if bundle := t.Bundle(); bundle != nil {
 		return bundle.TotalEth
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -522,18 +522,16 @@ func (t *TxWithMinerFee) SBundle() *SimSBundle {
 }
 
 func (t *TxWithMinerFee) Price() *big.Int {
-	return t.minerFee
+	return new(big.Int).Set(t.minerFee)
 }
 
 func (t *TxWithMinerFee) Profit() *big.Int {
-	if tx := t.Tx(); tx != nil {
-		return tx.Value()
-	} else if bundle := t.Bundle(); bundle != nil {
+	if bundle := t.Bundle(); bundle != nil {
 		return bundle.TotalEth
 	} else if sbundle := t.SBundle(); sbundle != nil {
 		return sbundle.Profit
 	} else {
-		return big.NewInt(0)
+		panic("profit called on unsupported order type")
 	}
 }
 
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -525,6 +525,18 @@ func (t *TxWithMinerFee) Price() *big.Int {
 	return t.minerFee
 }
 
+func (t *TxWithMinerFee) Profit() *big.Int {
+	if tx := t.Tx(); tx != nil {
+		return tx.Value()
+	} else if bundle := t.Bundle(); bundle != nil {
+		return bundle.TotalEth
+	} else if sbundle := t.SBundle(); sbundle != nil {
+		return sbundle.Profit
+	} else {
+		return big.NewInt(0)
+	}
+}
+
 // NewTxWithMinerFee creates a wrapped transaction, calculating the effective
 // miner gasTipCap if a base fee is provided.
 // Returns error in case of a negative effective miner gasTipCap.
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -521,6 +521,10 @@ func (t *TxWithMinerFee) SBundle() *SimSBundle {
 	return t.order.AsSBundle()
 }
 
+func (t *TxWithMinerFee) Price() *big.Int {
+	return t.minerFee
+}
+
 // NewTxWithMinerFee creates a wrapped transaction, calculating the effective
 // miner gasTipCap if a base fee is provided.
 // Returns error in case of a negative effective miner gasTipCap.
@@ -536,7 +540,7 @@ func NewTxWithMinerFee(tx *Transaction, baseFee *big.Int) (*TxWithMinerFee, erro
 }
 
 // NewBundleWithMinerFee creates a wrapped bundle.
-func NewBundleWithMinerFee(bundle *SimulatedBundle, baseFee *big.Int) (*TxWithMinerFee, error) {
+func NewBundleWithMinerFee(bundle *SimulatedBundle, _ *big.Int) (*TxWithMinerFee, error) {
 	minerFee := bundle.MevGasPrice
 	return &TxWithMinerFee{
 		order:    _BundleOrder{bundle},
@@ -545,7 +549,7 @@ func NewBundleWithMinerFee(bundle *SimulatedBundle, baseFee *big.Int) (*TxWithMi
 }
 
 // NewSBundleWithMinerFee creates a wrapped bundle.
-func NewSBundleWithMinerFee(sbundle *SimSBundle, baseFee *big.Int) (*TxWithMinerFee, error) {
+func NewSBundleWithMinerFee(sbundle *SimSBundle, _ *big.Int) (*TxWithMinerFee, error) {
 	minerFee := sbundle.MevGasPrice
 	return &TxWithMinerFee{
 		order:    _SBundleOrder{sbundle},
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -687,6 +687,12 @@ func (t *TransactionsByPriceAndNonce) Shift() {
 	heap.Pop(&t.heads)
 }
 
+// ShiftAndPushByAccountForTx attempts to update the transaction list associated with a given account address
+// based on the input transaction account. If the associated account exists and has additional transactions,
+// the top of the transaction list is popped and pushed to the heap.
+// Note that this operation should only be performed when the head transaction on the heap is different from the
+// input transaction. This operation is useful in scenarios where the current best head transaction for an account
+// was already popped from the heap and we want to process the next one from the same account.
 func (t *TransactionsByPriceAndNonce) ShiftAndPushByAccountForTx(tx *Transaction) {
 	if tx == nil {
 		return
--- a/miner/algo_common.go
+++ b/miner/algo_common.go
@@ -223,7 +223,7 @@ func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData
 }
 
 // Commit Bundle to env diff
-func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chData chainData, interrupt *int32, validation algorithmConfig) error {
+func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chData chainData, interrupt *int32, algoConf algorithmConfig) error {
 	coinbase := envDiff.baseEnvironment.coinbase
 	tmpEnvDiff := envDiff.copy()
 
@@ -297,14 +297,14 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
 		return errors.New("bundle underpays")
 	}
 
-	if validation.EnforceProfit {
+	if algoConf.EnforceProfit {
 		// if profit is enforced between simulation and actual commit, only allow ProfitThresholdPercent divergence
 		simulatedBundleProfit := new(big.Int).Set(bundle.TotalEth)
 		actualBundleProfit := new(big.Int).Mul(bundleActualEffGP, big.NewInt(int64(gasUsed)))
 
 		// We want to make simulated profit smaller to allow for some leeway in cases where the actual profit is
 		// lower due to transaction ordering
-		simulatedBundleProfit.Mul(simulatedBundleProfit, validation.ProfitThresholdPercent)
+		simulatedBundleProfit.Mul(simulatedBundleProfit, algoConf.ProfitThresholdPercent)
 		actualBundleProfit.Mul(actualBundleProfit, common.Big100)
 
 		if simulatedBundleProfit.Cmp(actualBundleProfit) > 0 {
@@ -436,7 +436,7 @@ func (envDiff *environmentDiff) commitPayoutTx(amount *big.Int, sender, receiver
 	return receipt, nil
 }
 
-func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainData, interrupt *int32, key *ecdsa.PrivateKey, validation algorithmConfig) error {
+func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainData, interrupt *int32, key *ecdsa.PrivateKey, algoConf algorithmConfig) error {
 	if key == nil {
 		return errors.New("no private key provided")
 	}
@@ -471,14 +471,14 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
 		return fmt.Errorf("incorrect EGP: got %d, expected %d", gotEGP, simEGP)
 	}
 
-	if validation.EnforceProfit {
+	if algoConf.EnforceProfit {
 		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
 		simulatedProfit := new(big.Int).Set(b.Profit)
 		actualProfit := new(big.Int).Set(coinbaseDelta)
 
 		// We want to make simulated profit smaller to allow for some leeway in cases where the actual profit is
 		// lower due to transaction ordering
-		simulatedProfit.Mul(simulatedProfit, validation.ProfitThresholdPercent)
+		simulatedProfit.Mul(simulatedProfit, algoConf.ProfitThresholdPercent)
 		actualProfit.Mul(actualProfit, common.Big100)
 
 		if simulatedProfit.Cmp(actualProfit) > 0 {
--- a/miner/algo_common.go
+++ b/miner/algo_common.go
@@ -39,6 +39,22 @@ var emptyCodeHash = common.HexToHash("c5d2460186f7233c927e7db2dcc703c0e500b653ca
 
 var errInterrupt = errors.New("miner worker interrupted")
 
+// lowProfitError is returned when an order is not committed due to low profit or low effective gas price
+type lowProfitError struct {
+	ExpectedProfit *big.Int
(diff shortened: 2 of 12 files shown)


miner/algo_common.go
```diff
@@ -262,14 +262,29 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
 
 	if bundleSimEffGP.Cmp(bundleActualEffGP) == 1 {
 		log.Trace("Bundle underpays after inclusion", "bundle", bundle.OriginalBundle.Hash)
 		return errors.New("bundle underpays")
 	}
 
+	if enforceProfit {
+		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
+		simulatedBundleProfit := bundle.TotalEth
+		actualBundleGasFees := new(big.Int).Mul(bundleActualEffGP, new(big.Int).SetUint64(gasUsed))
```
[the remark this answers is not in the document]
Wazzymandias: `gasUsed` is type `uint64` - i'm worried about loss of precision if I cast it

miner/algo_common.go
```diff
@@ -428,6 +443,21 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
 		return fmt.Errorf("incorrect EGP: got %d, expected %d", gotEGP, simEGP)
 	}
 
+	if enforceProfit {
```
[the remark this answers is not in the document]
Wazzymandias: This one is for `sbundle` rather than `bundle` - should they be calculated differently?

cc: @dvush

core/types/transaction.go
```diff
@@ -521,6 +521,22 @@ func (t *TxWithMinerFee) SBundle() *SimSBundle {
...
+
+func (t *TxWithMinerFee) Profit() *big.Int {
+	if tx := t.Tx(); tx != nil {
+		return tx.Value()
+	} else if bundle := t.Bundle(); bundle != nil {
+		return bundle.TotalEth
+	} else if sbundle := t.SBundle(); sbundle != nil {
+		return sbundle.Profit
+	} else {
+		return big.NewInt(0)
```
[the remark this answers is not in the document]
Wazzymandias: agreed, updated to panic

core/types/transaction.go
```diff
@@ -521,6 +521,22 @@ func (t *TxWithMinerFee) SBundle() *SimSBundle {
 	return t.order.AsSBundle()
 }
 
+func (t *TxWithMinerFee) Price() *big.Int {
+	return t.minerFee
```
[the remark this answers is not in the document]
Wazzymandias: good catch! updated to copy the object

miner/algo_greedy.go
```diff
@@ -20,21 +22,154 @@ type greedyBuilder struct {
 	chainData        chainData
 	builderKey       *ecdsa.PrivateKey
 	interrupt        *int32
+	algoType         AlgoType
```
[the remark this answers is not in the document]
Wazzymandias: makes sense, agreed - created separate greedy buckets builder

miner/algo_greedy.go
```diff
@@ -20,21 +22,154 @@ type greedyBuilder struct {
...
+		}
+
+		SortByProfit = func(transactions []*types.TxWithMinerFee) {
+			sort.SliceStable(transactions, func(i, j int) bool {
+				var (
+					iProfit = transactions[i].Profit()
+					jProfit = transactions[j].Profit()
+				)
+
+				return iProfit.Cmp(jProfit) > 0
```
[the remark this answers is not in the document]
Wazzymandias: updated

miner/algo_greedy.go
```diff
@@ -20,21 +22,154 @@ type greedyBuilder struct {
...
+		return nil, nil
+	}
+
+	var (
+		usedBundles       []types.SimulatedBundle
+		usedSbundles      []types.UsedSBundle
+		transactionBucket []*types.TxWithMinerFee
+		percent           = big.NewFloat(0.9)
+
+		InitializeBucket = func(order *types.TxWithMinerFee) [1]*big.Int {
```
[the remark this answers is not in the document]
Wazzymandias: good point, helped with initial reasoning but updated to just consider cutoff price

miner/algo_greedy.go
```diff
@@ -20,21 +22,154 @@ type greedyBuilder struct {
...
+				usedBundles = append(usedBundles, bundles...)
+				usedSbundles = append(usedSbundles, sbundles...)
+				transactionBucket = nil
+				continue // re-run since committing transactions may have pushed higher nonce transactions back into heap
+			}
+			// TODO: don't break if there are still retryable transactions
+			break
+		}
+
+		if ok := IsOrderInPriceRange(order, bucket[0]); ok {
```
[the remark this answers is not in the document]
Wazzymandias: agreed, updated to use cutoff price instead

miner/algo_greedy.go
```diff
@@ -20,21 +22,154 @@ type greedyBuilder struct {
...
+			floorPrice := new(big.Float).Mul(new(big.Float).SetInt(order.Price()), percent)
+			round, _ := floorPrice.Int64()
+			return [1]*big.Int{big.NewInt(round)}
+		}
+
+		IsOrderInPriceRange = func(order *types.TxWithMinerFee, minPrice *big.Int) bool {
+			return order.Price().Cmp(minPrice) > -1
+		}
+
+		SortByProfit = func(transactions []*types.TxWithMinerFee) {
```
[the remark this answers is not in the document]
Wazzymandias: updated to `SortInPlaceByProfit`

dvush (commented):
I think this is what we want!
+ left some comment

miner/algo_common.go
```diff
@@ -428,6 +443,21 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
...
+		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
+		simulatedProfit := b.Profit
+		actualGasFees := new(big.Int).Mul(gasDelta, gotEGP)
+		actualProfit := new(big.Int).Add(coinbaseDelta, actualGasFees)
+
```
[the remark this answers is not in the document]
Wazzymandias: good point, updated, please feel free to un-resolve if there's still concerns

miner/algo_common.go
```diff
@@ -428,6 +443,21 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
...
+		simulatedProfit.Mul(simulatedProfit, common.Big100)
+		actualProfit.Mul(actualProfit, big.NewInt(99))
```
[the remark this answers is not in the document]
Wazzymandias: good point, updated, please feel free to un-resolve if there's still concerns

miner/algo_common.go
```diff
@@ -262,14 +262,29 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
 	}
 
+	if enforceProfit {
+		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
+		simulatedBundleProfit := bundle.TotalEth
+		actualBundleGasFees := new(big.Int).Mul(bundleActualEffGP, big.NewInt(int64(gasUsed)))
+		actualBundleProfit := new(big.Int).Add(coinbaseBalanceDelta, actualBundleGasFees)
+
+		simulatedBundleProfit.Mul(simulatedBundleProfit, common.Big100)
+		actualBundleProfit.Mul(actualBundleProfit, big.NewInt(99))
```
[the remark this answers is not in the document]
Wazzymandias: good point, updated, please feel free to un-resolve if there's still concerns

core/types/transaction.go
```diff
@@ -521,6 +521,24 @@ func (t *TxWithMinerFee) SBundle() *SimSBundle {
...
 }
 
+func (t *TxWithMinerFee) Price() *big.Int {
+	return new(big.Int).Set(t.minerFee)
+}
+
+func (t *TxWithMinerFee) Profit(baseFee *big.Int) *big.Int {
+	if tx := t.Tx(); tx != nil {
+		profit := new(big.Int).Sub(tx.GasPrice(), baseFee)
+		profit.Mul(profit, new(big.Int).SetUint64(tx.Gas()))
```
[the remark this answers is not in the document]
Wazzymandias: Do you have suggestion for what we could do instead?

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+				Mul(
+					new(big.Float).SetInt(order.Price()),
+					percent,
+				)
+			round, _ := floorPrice.Int64()
+			return big.NewInt(round)
+		}
+
+		IsOrderInPriceRange = func(order *types.TxWithMinerFee, minPrice *big.Int) bool {
+			return order.Price().Cmp(minPrice) > -1
```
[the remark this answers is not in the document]
Wazzymandias: agreed, updated, please feel free to un-resolve if there are still concerns

core/types/transaction.go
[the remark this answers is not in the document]
Wazzymandias: cc: @Ruteri @avalonche

miner/algo_common.go
```diff
@@ -262,14 +262,29 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
 	if bundleSimEffGP.Cmp(bundleActualEffGP) == 1 {
 		log.Trace("Bundle underpays after inclusion", "bundle", bundle.OriginalBundle.Hash)
 		return errors.New("bundle underpays")
 	}
 
+	if enforceProfit {
+		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
+		simulatedBundleProfit := bundle.TotalEth
+		actualBundleGasFees := new(big.Int).Mul(bundleActualEffGP, big.NewInt(int64(gasUsed)))
+		actualBundleProfit := new(big.Int).Add(coinbaseBalanceDelta, actualBundleGasFees)
```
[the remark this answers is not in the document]
Wazzymandias: there was a unit test that failed when i tried your suggestion. i've made a modification that i think addresses your concern. please feel free to un-resolve or create another comment if there are still issues

core/types/transaction.go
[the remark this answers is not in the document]
Ruteri: What I can suggest is to re-insert the tx after its apply to the block if its total profit (or the egp) changes enough to move it to the next slot by more than some 1%
Simulating txs up front is also on the table I guess

miner/algo_common.go
```diff
@@ -262,14 +262,31 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
+		simulatedBundleProfit := new(big.Int).Set(bundle.TotalEth)
+		actualBundleGasFees := new(big.Int).Mul(bundleActualEffGP, big.NewInt(int64(gasUsed)))
+		actualBundleProfit := new(big.Int).Add(coinbaseBalanceDelta, actualBundleGasFees)
+
+		// We want to make simulated profit smaller to allow for some leeway in cases where the actual profit is
+		// lower due to transaction ordering
+		simulatedBundleProfit.Mul(simulatedBundleProfit, big.NewInt(99))
+		actualBundleProfit.Mul(actualBundleProfit, common.Big100)
+
+		if simulatedBundleProfit.Cmp(actualBundleProfit) > 0 {
```
Ruteri, line 284: I think that the 99% cutoff is very aggressive for profit, especially that for regular txs you don't take into account unused gas

miner/algo_common.go
```diff
@@ -428,6 +445,22 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
 		return fmt.Errorf("incorrect EGP: got %d, expected %d", gotEGP, simEGP)
 	}
 
+	if enforceProfit {
```
Ruteri, line 448: Q: why do you have a separate flag that's not present for EGP check? Is it only for testing?

miner/algo_greedy.go
```diff
@@ -31,10 +34,12 @@ func newGreedyBuilder(chain *core.BlockChain, chainConfig *params.ChainConfig, b
 	}
 }
 
-func (b *greedyBuilder) mergeOrdersIntoEnvDiff(envDiff *environmentDiff, orders *types.TransactionsByPriceAndNonce) ([]types.SimulatedBundle, []types.UsedSBundle) {
-	usedBundles := []types.SimulatedBundle{}
-	usedSbundles := []types.UsedSBundle{}
-
+func (b *greedyBuilder) mergeOrdersIntoEnvDiff(
```
Ruteri, line 37: I find this formatting confusing, it's not present elsewhere in the codebase - I'd either stick to the codebase's style, or make the formatting a bit more obvious
```
func (b *greedyBuilder) mergeOrdersIntoEnvDiff(
	envDiff *environmentDiff, orders *types.TransactionsByPriceAndNonce) ([]types.SimulatedBundle, []types.UsedSBundle
) {
```

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+func (b *greedyBucketsBuilder) commit(envDiff *environmentDiff,
+	transactions []*types.TxWithMinerFee,
+	orders *types.TransactionsByPriceAndNonce,
+	retryMap map[*types.TxWithMinerFee]int, retryLimit int,
+) ([]types.SimulatedBundle, []types.UsedSBundle) {
+	var (
+		usedBundles                []types.SimulatedBundle
+		usedSbundles               []types.UsedSBundle
+		CheckRetryOrderAndReinsert = func(order *types.TxWithMinerFee, orders *types.TransactionsByPriceAndNonce, retryMap map[*types.TxWithMinerFee]int, retryLimit int) {
+			var isRetryable bool
```
Ruteri, line 48: suggestion: make it explicit `var isRetryable bool = false`

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+				usedEntry.Success = false
+				usedSbundles = append(usedSbundles, usedEntry)
+				CheckRetryOrderAndReinsert(order, orders, retryMap, retryLimit)
+				continue
+			}
+
+			log.Trace("Included sbundle", "bundleEGP", sbundle.MevGasPrice.String(), "ethToCoinbase", ethIntToFloat(sbundle.Profit))
+			usedEntry.Success = true
+			usedSbundles = append(usedSbundles, usedEntry)
+		}
```
Ruteri, line 110: suggestion: else panic / return error

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, nil
+	}
+
+	const (
+		retryLimit = 2
+	)
+
+	var (
+		baseFee      = envDiff.baseEnvironment.header.BaseFee
+		retryMap     = make(map[*types.TxWithMinerFee]int)
```
Ruteri, line 127: suggestion: if you are anticipating retries, it'd probably be better to pre-allocate for len(orders)

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (b *greedyBucketsBuilder) mergeOrdersIntoEnvDiff(
+	envDiff *environmentDiff, orders *types.TransactionsByPriceAndNonce) ([]types.SimulatedBundle, []types.UsedSBundle) {
+	if orders.Peek() == nil {
+		return nil, nil
+	}
+
+	const (
+		retryLimit = 2
```
Ruteri, line 122: This seems huge, since most orders won't "get fixed", and most orders will fail.
I'd probably suggest only retrying once, and only if the error is "bundle underpays"

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (b *greedyBucketsBuilder) commit(envDiff *environmentDiff,
+	transactions []*types.TxWithMinerFee,
+	orders *types.TransactionsByPriceAndNonce,
+	retryMap map[*types.TxWithMinerFee]int, retryLimit int,
+) ([]types.SimulatedBundle, []types.UsedSBundle) {
+	var (
+		usedBundles                []types.SimulatedBundle
+		usedSbundles               []types.UsedSBundle
+		CheckRetryOrderAndReinsert = func(order *types.TxWithMinerFee, orders *types.TransactionsByPriceAndNonce, retryMap map[*types.TxWithMinerFee]int, retryLimit int) {
```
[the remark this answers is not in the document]
Wazzymandias: it's only used in one place - if it ends up being used elsewhere it'd be pretty easy to move it out

miner/algo_greedy_buckets.go
Wazzymandias: there's no `len` operation on orders unfortunately - it's a container for the underlying heap and doesn't expose access to the heap slice

miner/algo_greedy.go
Wazzymandias: fair, updated

miner/algo_greedy_buckets.go
Wazzymandias: added

miner/algo_common.go
Wazzymandias: @dvush suggested adding enforce profit flag for backwards compatibility with algo greedy builder

core/types/transaction.go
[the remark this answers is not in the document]
Wazzymandias: > Simulating txs up front is also on the table

For the scope of this PR, I'd like to avoid simulating the transaction due to the performance hit. 

> Txs will be a bit abusable here.

I agree. I spoke with Bert and we think that this could be mitigated by leveraging the broader builder network and constraining the subset of builders that use this algorithm. Open to other ideas as well @dvush @Ruteri @avalonche

miner/algo_common.go
Wazzymandias: 99% was recommendation from @dvush 

From a spot check analysis it appears rare for low profit to occur

miner/algo_greedy_buckets.go
Wazzymandias: updated retry limit

miner/algo_common.go
```diff
@@ -262,14 +262,31 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
 	if bundleSimEffGP.Cmp(bundleActualEffGP) == 1 {
 		log.Trace("Bundle underpays after inclusion", "bundle", bundle.OriginalBundle.Hash)
 		return errors.New("bundle underpays")
 	}
 
+	if enforceProfit {
+		// if profit is enforced between simulation and actual commit, only allow >-1% divergence
+		simulatedBundleProfit := new(big.Int).Set(bundle.TotalEth)
+		actualBundleGasFees := new(big.Int).Mul(bundleActualEffGP, big.NewInt(int64(gasUsed)))
+		actualBundleProfit := new(big.Int).Add(coinbaseBalanceDelta, actualBundleGasFees)
```
dvush, line 277: its still looks incorrect. coinbaseBalanceDelta is actual bundle profit

miner/algo_common.go
dvush: @Ruteri enforcing profit is important when you use profit to actually sort bundles. Its not needed for greedy builder because it uses EGP for ordering. If you don't enforce, then you can simply do the same thing that was done with EGP manipulation before EGP check was introduced.

miner/algo_common.go
dvush: I don't know what cutoff is better here. 5%?

core/types/transaction.go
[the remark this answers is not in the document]
dvush: It can work because we can be sure that MEV is in the bundles but not txs so txs should not conflict with bundles that much. 

But its also can be wrong because it will prioritize mempool swaps over bundles. Here is univ2 router txs https://etherscan.io/txs?a=0x7a250d5630b4cf539739df2c5dacb4c659f2488d gas limit is much bigger than gas used.

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,211 @@
...
+
+		if ok := IsOrderInPriceRange(order, minPrice); ok {
+			orders.Pop()
+			transactions = append(transactions, order)
+		} else {
+			if len(transactions) != 0 {
+				SortInPlaceByProfit(baseFee, transactions, gasUsedMap)
+				bundles, sbundles := b.commit(envDiff, transactions, orders, gasUsedMap, retryMap, retryLimit)
+				usedBundles = append(usedBundles, bundles...)
+				usedSbundles = append(usedSbundles, sbundles...)
```
dvush, line 195: do you handle retry correctly? there is an assumption that you will return used bundles where one bundle will be there once

core/types/transaction.go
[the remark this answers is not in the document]
Wazzymandias: I updated to use actual gas used on reinsert

miner/algo_common.go
Wazzymandias: updated to 5%

miner/algo_common.go
Wazzymandias: fixed, discussed offline

miner/algo_greedy_buckets.go
Wazzymandias: we only insert sbundle after reinsert fails such that sbundle is only appended once

miner/algo_common.go
```diff
@@ -264,14 +264,30 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
 	bundleSimEffGP := new(big.Int).Set(bundle.MevGasPrice)
 
 	// allow >-1% divergence
-	bundleActualEffGP.Mul(bundleActualEffGP, big.NewInt(100))
+	bundleActualEffGP.Mul(bundleActualEffGP, common.Big100)
```
Ruteri, line 267: Should this check be done if `enforceProfit` is set?

miner/algo_common.go
Ruteri: I'm suggesting to perform both checks still, just outside of this function (by greedy and buckets respectively)

miner/algo_common.go
```diff
@@ -163,31 +163,31 @@ func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData
...
+			return receipt, popTx, err
 
 		default:
 			// Strange error, discard the transaction and get the next in line (note, the
 			// nonce-too-high clause will prevent us from executing in vain).
 			log.Trace("Transaction failed, account skipped", "hash", tx.Hash(), "err", err)
-			return nil, shiftTx, err
+			return receipt, shiftTx, err
 		}
 	}
```
Ruteri, line 193: Do we miss profit enforcement check here?

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,213 @@
...
+			return big.NewInt(round)
+		}
+
+		IsOrderInPriceRange = func(order *types.TxWithMinerFee, minPrice *big.Int) bool {
+			return order.Price().Cmp(minPrice) >= 0
+		}
+
+		SortInPlaceByProfit = func(baseFee *big.Int, transactions []*types.TxWithMinerFee, gasUsedMap map[*types.TxWithMinerFee]uint64) {
+			sort.SliceStable(transactions, func(i, j int) bool {
+				return transactions[i].Profit(baseFee, gasUsedMap[transactions[i]]).Cmp(transactions[j].Profit(baseFee, gasUsedMap[transactions[j]])) > 0
```
Ruteri, line 167: isn't gasUsedMap all zeroes?

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,213 @@
...
+	if orders.Peek() == nil {
+		return nil, nil
+	}
+
+	const retryLimit = 1
+
+	var (
+		baseFee      = envDiff.baseEnvironment.header.BaseFee
+		retryMap     = make(map[*types.TxWithMinerFee]int)
+		gasUsedMap   = make(map[*types.TxWithMinerFee]uint64)
```
Ruteri, line 145: Would it be better to have a `gasUseOverrideMap`, only updating the items that actually did not apply correctly?

miner/algo_common.go
Wazzymandias: added profit enforcement check for tx

miner/algo_common.go
Wazzymandias: this is EGP enforcement, not necessarily profit enforcement

miner/algo_common.go
```diff
@@ -22,10 +22,33 @@ const (
...
+		ExpectedProfit:         common.Big0,
+		ProfitThresholdPercent: defaultProfitThreshold,
+	}
+)
+
 var emptyCodeHash = common.HexToHash("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
 
 var errInterrupt = errors.New("miner worker interrupted")
 
+type validationConfig struct {
```
Ruteri, line 42: I think the name is a bit misleading, maybe algo config? apply config?

miner/algo_common.go
```diff
@@ -156,49 +179,63 @@ func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData
...
 		}
 	}
 
 	envDiff.newProfit = envDiff.newProfit.Add(envDiff.newProfit, gasPrice.Mul(gasPrice, big.NewInt(int64(receipt.GasUsed))))
 	envDiff.newTxs = append(envDiff.newTxs, tx)
 	envDiff.newReceipts = append(envDiff.newReceipts, receipt)
+
+	if validation.EnforceProfit {
+		if validation.ExpectedProfit == nil {
+			return receipt, shiftTx, errors.New("expected profit is nil")
```
Ruteri, line 224: hm, the results were added to the `envDiff` already though, right? Along with the new state

miner/algo_common.go
Wazzymandias: updated to `algorithmConfig`

miner/algo_common.go
Wazzymandias: discussed offline - rolling back enforcement on txs for now

miner/worker.go
```diff
@@ -1368,40 +1370,162 @@ func (w *worker) fillTransactions(interrupt *int32, env *environment) (error, []
...
+		newEnv       *environment
+		blockBundles []types.SimulatedBundle
+		usedSbundle  []types.UsedSBundle
+	)
+
 	start := time.Now()
-	newEnv, blockBundles, usedSbundle := builder.buildBlock(bundlesToConsider, sbundlesToConsider, pending)
+
+	switch w.flashbots.algoType {
+	case ALGO_TEST_ONLY:
```
Ruteri, line 1407: Do we want to merge this code too? I'd probably prefer it lives on a branch

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,228 @@
...
+		retryMap     = make(map[*types.TxWithMinerFee]int)
+		usedBundles  []types.SimulatedBundle
+		usedSbundles []types.UsedSBundle
+		transactions []*types.TxWithMinerFee
+		percent      = new(big.Float).Quo(
+			new(big.Float).SetInt(b.algoConf.ProfitThresholdPercent),
+			new(big.Float).SetInt(common.Big100),
+		)
+
+		CutoffPriceFromOrder = func(order *types.TxWithMinerFee, cutoffPercent *big.Float) *big.Int {
```
Ruteri, line 166: Do you maybe want to move the closures into regular standalone functions for better readability?

Ruteri (approved):

miner/worker.go
Wazzymandias: agreed, removed

miner/algo_greedy_buckets.go
```diff
@@ -0,0 +1,228 @@
...
+	transactions []*types.TxWithMinerFee,
+	orders *types.TransactionsByPriceAndNonce,
+	gasUsedMap map[*types.TxWithMinerFee]uint64, retryMap map[*types.TxWithMinerFee]int, retryLimit int,
+) ([]types.SimulatedBundle, []types.UsedSBundle) {
+	var (
+		usedBundles  []types.SimulatedBundle
+		usedSbundles []types.UsedSBundle
+		algoConf     = b.algoConf
+
+		CheckRetryOrderAndReinsert = func(
```
dvush, line 60: I think this kind of retry is likely to be useless.

It will actually never work in the case that make sense - when bundle is inserted successfully but we fail it because it underpays. For it to work we need to update bundle mev gas price and profit with values that bundle got when it failed and reinsert it with a new values to the heap. 

No need to change anything, just an opinion.

dvush (approved):

miner/algo_greedy_buckets.go
Wazzymandias: Updated to only reinsert when error is due to low profit, and added logic to update profit and price prior to reinserting to heap

miner/algo_greedy_buckets.go
Wazzymandias: moved some of the closures into standalone functions

PR #78, opened 2023-06-20
Allow alternative methods of proposer payments in validation api.

dvush:
## 📝 Summary

This will allow validation API to accept blocks where:
* proposer payment is plain transfer to the fee recipient
* proposer payment is internal payment to the fee recipient made through the smart contract
* fee recipient is set to coinbase and accepts payment directly

The way we calculate profit is:
* check balance difference of the fee recipient after each tx in the block
* if its positive we count it toward total block profit
* if its negative we ignore it, but if fee recipient is also a coinbase we add tip payed by the tx to profit
* when profit is calculated we make sure that claimed profit is equal or bigger than calculated profit

Some features of this method:
* withdrawals from the fee recipient don't decrease profit of the bid
* deposits to the fee recipient address can be claimed as a profit
* builders that ignore withdrawals/deposits of the fee recipients and just calculate balance diff of it before and after block will work.  

---

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -40,6 +40,7 @@ import (
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/utils"
 	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/eth/tracers/logger"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/internal/syncx"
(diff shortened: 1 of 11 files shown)


metachris (approved):
would love a review from @Ruteri too

Wazzymandias (approved):

core/blockchain.go
```diff
@@ -2554,6 +2568,44 @@ func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Ad
...
+					if err != nil {
+						log.Error("failed to calculate tip", "err", err)
+						break
+					}
+					profit := tip.Mul(tip, new(big.Int).SetUint64(receipts[i].GasUsed))
+					feeRecipientProfit.Add(feeRecipientProfit, profit)
+				}
+			}
+		}
+		if feeRecipientProfit.Cmp(expectedProfit) >= 0 {
```
Ruteri, line 2602: Hm, maybe `== 0` would help us notice bugs faster

dvush: >= here is so that builders can underestimate their profits. e.g. if you don't count payments to the proposals and withdrawals correctly.

dvush:
closed in favor of https://github.com/flashbots/builder/pull/93

PR #81, opened 2023-07-02
Toggleable discard for failed revertible transaction hashes 

Wazzymandias:
## 📝 Summary

- Current behavior of bundles is that:
  1. All transactions in a bundle must be included in specified order, with no transaction inserted between them, **OR** 
  2. If any transaction reverts, then no transaction from the bundle will be included **EXCEPT**
  3. When user specifies **reverting transaction hashes** in their request, the builder can still include the bundle with one or more of the transactions failing within it

- This PR makes it so when users specify reverting transaction hashes in their request, the builder **discards** the transaction(s) on **error**

### Changes 
- Add toggleable support for discarding reverted transaction hashes on error
- When a transaction that's part of a bundle fails, and the request sent to the builder sets field to allow revert for the transaction hash, the builder will discard the hash of the failed transaction from the submitted bundle. 
  - For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition 

## 📚 References

- https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition

---

### DIFF
--- a/builder/config.go
+++ b/builder/config.go
@@ -25,6 +25,7 @@ type Config struct {
 	BuilderRateLimitMaxBurst         int           `toml:",omitempty"`
 	BuilderRateLimitResubmitInterval string        `toml:",omitempty"`
 	BuilderSubmissionOffset          time.Duration `toml:",omitempty"`
+	DiscardRevertedHashes            bool          `toml:",omitempty"`
 	EnableCancellations              bool          `toml:",omitempty"`
 }
 
--- a/builder/config.go
+++ b/builder/config.go
@@ -51,6 +51,7 @@ var DefaultConfig = Config{
 	ValidationBlocklist:           "",
 	BuilderRateLimitDuration:      RateLimitIntervalDefault.String(),
 	BuilderRateLimitMaxBurst:      RateLimitBurstDefault,
+	DiscardRevertedHashes:         false,
 	EnableCancellations:           false,
 }
 
--- a/builder/config.go
+++ b/builder/config.go
@@ -25,7 +25,7 @@ type Config struct {
 	BuilderRateLimitMaxBurst         int           `toml:",omitempty"`
 	BuilderRateLimitResubmitInterval string        `toml:",omitempty"`
 	BuilderSubmissionOffset          time.Duration `toml:",omitempty"`
-	DiscardRevertedHashes            bool          `toml:",omitempty"`
+	DiscardRevertibleTxOnErr         bool          `toml:",omitempty"`
 	EnableCancellations              bool          `toml:",omitempty"`
 }
 
@@ -51,7 +51,7 @@ var DefaultConfig = Config{
 	ValidationBlocklist:           "",
 	BuilderRateLimitDuration:      RateLimitIntervalDefault.String(),
 	BuilderRateLimitMaxBurst:      RateLimitBurstDefault,
-	DiscardRevertedHashes:         false,
+	DiscardRevertibleTxOnErr:      false,
 	EnableCancellations:           false,
 }
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -850,6 +850,9 @@ var (
 		Usage: "When enabled, if a transaction submitted as part of a bundle in a send bundle request is reverted, " +
 			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the reverted transaction from the submitted bundle." +
 			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
+		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTED_HASHES"},
+		Value:    builder.DiscardRevertedHashesDefault,
+		Category: flags.BuilderCategory,
 	}
 
 	BuilderEnableCancellations = &cli.BoolFlag{
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -845,6 +845,13 @@ var (
 		Category: flags.BuilderCategory,
 	}
 
+	BuilderDiscardRevertedHashes = &cli.BoolFlag{
+		Name: "builder.discard_reverted_hashes",
+		Usage: "When enabled, if a transaction submitted as part of a bundle in a send bundle request is reverted, " +
+			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the reverted transaction from the submitted bundle." +
+			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
+	}
+
 	BuilderEnableCancellations = &cli.BoolFlag{
 		Name:     "builder.cancellations",
 		Usage:    "Enable cancellations for the builder",
@@ -1668,6 +1675,7 @@ func SetBuilderConfig(ctx *cli.Context, cfg *builder.Config) {
 	cfg.BuilderRateLimitDuration = ctx.String(BuilderRateLimitDuration.Name)
 	cfg.BuilderRateLimitMaxBurst = ctx.Int(BuilderRateLimitMaxBurst.Name)
 	cfg.BuilderSubmissionOffset = ctx.Duration(BuilderSubmissionOffset.Name)
+	cfg.DiscardRevertedHashes = ctx.Bool(BuilderDiscardRevertedHashes.Name)
 	cfg.EnableCancellations = ctx.IsSet(BuilderEnableCancellations.Name)
 }
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -845,10 +845,10 @@ var (
 		Category: flags.BuilderCategory,
 	}
 
-	BuilderDiscardRevertedHashes = &cli.BoolFlag{
-		Name: "builder.discard_reverted_hashes",
-		Usage: "When enabled, if a transaction submitted as part of a bundle in a send bundle request is reverted, " +
-			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the reverted transaction from the submitted bundle." +
+	BuilderDiscardRevertibleTxOnErr = &cli.BoolFlag{
+		Name: "builder.discard_revertible_tx_on_error",
+		Usage: "When enabled, if a transaction submitted as part of a bundle in a send bundle request has error on commit, " +
+			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the failed transaction from the submitted bundle." +
 			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
 		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTED_HASHES"},
 		Category: flags.BuilderCategory,
@@ -1677,7 +1677,7 @@ func SetBuilderConfig(ctx *cli.Context, cfg *builder.Config) {
 	cfg.BuilderRateLimitDuration = ctx.String(BuilderRateLimitDuration.Name)
 	cfg.BuilderRateLimitMaxBurst = ctx.Int(BuilderRateLimitMaxBurst.Name)
 	cfg.BuilderSubmissionOffset = ctx.Duration(BuilderSubmissionOffset.Name)
-	cfg.DiscardRevertedHashes = ctx.Bool(BuilderDiscardRevertedHashes.Name)
+	cfg.DiscardRevertibleTxOnErr = ctx.Bool(BuilderDiscardRevertibleTxOnErr.Name)
 	cfg.EnableCancellations = ctx.IsSet(BuilderEnableCancellations.Name)
 }
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -850,7 +850,7 @@ var (
 		Usage: "When enabled, if a transaction submitted as part of a bundle in a send bundle request has error on commit, " +
 			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the failed transaction from the submitted bundle." +
 			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
-		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTED_HASHES"},
+		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTIBLE_TX_ON_ERROR"},
 		Category: flags.BuilderCategory,
 	}
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -877,6 +877,7 @@ var (
 			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the failed transaction from the submitted bundle." +
 			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
 		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTIBLE_TX_ON_ERROR"},
+		Value:    builder.DefaultConfig.DiscardRevertibleTxOnErr,
 		Category: flags.BuilderCategory,
 	}
 
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -851,7 +851,6 @@ var (
 			"and its hash is specified as one that can revert in the request body, the builder will discard the hash of the reverted transaction from the submitted bundle." +
 			"For additional details on the structure of the request body, see https://docs.flashbots.net/flashbots-mev-share/searchers/understanding-bundles#bundle-definition",
 		EnvVars:  []string{"FLASHBOTS_BUILDER_DISCARD_REVERTED_HASHES"},
-		Value:    builder.DiscardRevertedHashesDefault,
 		Category: flags.BuilderCategory,
 	}
 
--- a/miner/algo_common.go
+++ b/miner/algo_common.go
@@ -27,12 +27,10 @@ const (
 const defaultProfitPercentMinimum = 70
 
 var (
-	defaultProfitThreshold = big.NewInt(defaultProfitPercentMinimum)
 	defaultAlgorithmConfig = algorithmConfig{
 		DropTransactionOnRevert: false,
 		EnforceProfit:           false,
-		ExpectedProfit:          common.Big0,
-		ProfitThresholdPercent:  defaultProfitThreshold,
+		ProfitThresholdPercent:  defaultProfitPercentMinimum,
 	}
 )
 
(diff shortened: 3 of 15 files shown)


miner/algo_common.go
```diff
@@ -181,21 +185,40 @@ func applyTransactionWithBlacklist(signer types.Signer, config *params.ChainConf
...
 		envDiff.gasPool, envDiff.state, header, tx, &header.GasUsed, *chData.chain.GetVMConfig(), chData.blacklist)
 
+	// when drop transaction is enabled:
+	// 1. if there was an error applying the transaction OR
+	// 2. if the transaction receipt status is failed
+	// we don't apply the transaction to the state and we don't add it to the newTxs, return early
+	if algoConf.DropTransactionOnRevert && (err != nil || (receipt != nil && receipt.Status == types.ReceiptStatusFailed)) {
+		// the StateDB for environment diff is already modified at this point, since it gets mutated when passed in to
+		// applyTransactionWithBlacklist, so we need to revert it
+		s, sdbErr := state.New(root, envDiff.state.Database(), chData.chain.Snapshots())
```
dvush, line 212: I don't thing that this will work. If this bundle is in the middle of the block root of the state before it is unknown to the database and snapshots. 

And why original root is used here? By the name of it it looks like it will revert to the state before anything was applied to the state. 

Anyway, if you need to revert state to the OK state after tx failed to apply (nonce errror, etc) it can be done simpler. Just use statedb.Snapshot and revert.

And if you need to discard tx if it was applied successfully (but its probably not needed - see Mateusz comment above) it could be done without discarding statedb too. See how blocklist is implemented.

Wazzymandias: >  If this bundle is in the middle of the block root of the state before it is unknown to the database and snapshots

I don't think this is true because when we apply the intermediate state hash is calculated - I confirmed this in test as well

> Anyway, if you need to revert state to the OK state after tx failed to apply (nonce errror, etc) it can be done simpler. Just use statedb.Snapshot and revert.

I tried this approach initially but it failed unfortunately. Using snapshot revert was not working.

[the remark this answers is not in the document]
Wazzymandias: We only exclude when flag is enabled and transaction is in reverting transaction hashes list, logic can be found in `commitBundle`

```
		// if drop transaction on revert is enabled and the transaction is found in list of reverting transaction hashes,
		// we can consider the transaction for discarding
		canDiscard = discard && bundle.OriginalBundle.RevertingHash(tx.Hash())
		algoConf.DropTransactionOnRevert = canDiscard

```

Wazzymandias: @dvush this is error I get with approach `Just use statedb.Snapshot and revert`:

```
panic: revision id 0 cannot be reverted [recovered]
	panic: revision id 0 cannot be reverted
```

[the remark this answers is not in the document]
Wazzymandias: > why would we want to exclude reverting transactions though

This was the feature request - product requirement was that:
```
Update builder logic such that when reverting transaction hashes are specified, the builder discards transactions on revert for bundles and sbundles
```

builder/config.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	BuilderRateLimitMaxBurst         int           `toml:",omitempty"`
 	BuilderRateLimitResubmitInterval string        `toml:",omitempty"`
 	BuilderSubmissionOffset          time.Duration `toml:",omitempty"`
+	DiscardRevertedHashes            bool          `toml:",omitempty"`
```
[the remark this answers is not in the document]
Wazzymandias: updated to use `DefaultConfig`

miner/algo_common.go
```diff
@@ -181,21 +185,40 @@ func applyTransactionWithBlacklist(signer types.Signer, config *params.ChainConf
...
+		// if we cannot create a new state from the existing database, snapshots, and intermediate root hash,
+		// we panic because this is a fatal error
+		if sdbErr != nil {
+			panic(sdbErr)
```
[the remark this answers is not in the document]
Wazzymandias: updated

miner/algo_common.go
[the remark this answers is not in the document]
Wazzymandias: Perhaps I misunderstood. Updated the logic based on @Ruteri and @dvush feedback. 

> if it can't be included (e.g. nonce error) - we just skip it (new thing)

This is now what is implemented in this PR. Please un-resolve if there is still issue.

Wazzymandias: > final hash is calculated once

You are right, I was looking at pre-byzantium which calculates intermediate hash.

Since `applyTransactionWithBlacklist` already reverts on error, I have simplified the logic. Please un-resolve if there are still concerns.

miner/algo_common.go
```diff
@@ -551,14 +588,22 @@ func (envDiff *environmentDiff) commitSBundleInner(b *types.SBundle, chData chai
 
 		if el.Tx != nil {
 			receipt, _, err := envDiff.commitTx(el.Tx, chData)
+
 			if err != nil {
+				// if drop enabled, and revertible tx has error on commit,
+				// we skip the transaction and continue with next one
+				if algoConf.DropRevertibleTxOnErr && el.CanRevert {
+					log.Warn("Found error on commit for revertible tx, but discard on err is enabled so skipping.",
```
Ruteri, line 596: Suggestion: Warn -> Info since it's expected behaviour (we do want to log it tho I agree)

miner/algo_common.go
```diff
@@ -281,15 +296,28 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
+		}
+
+		if hasCommitErr {
+			log.Trace("Bundle tx error", "bundle", bundle.OriginalBundle.Hash, "tx", txHash, "err", err)
 			return err
 		}
 
-		if receipt.Status != types.ReceiptStatusSuccessful && !bundle.OriginalBundle.RevertingHash(tx.Hash()) {
-			log.Trace("Bundle tx failed", "bundle", bundle.OriginalBundle.Hash, "tx", tx.Hash(), "err", err)
+		if hasReceiptFailed && !isRevertibleTx {
```
Ruteri, line 319: From a readability perspective, I'd say putting `receipt != nil && receipt.Status == types.ReceiptStatusFailed` verbatim here is easier to read, since one has to jump around the code to see the definition of the variable.
Alternatively putting `hasReceiptFailed := receipt != nil && receipt.Status == types.ReceiptStatusFailed` right before this if would also be good

miner/algo_common.go
```diff
@@ -240,15 +247,23 @@ func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData
...
+
+		profitBefore = new(big.Int).Set(tmpEnvDiff.newProfit)
 
-	coinbaseBalanceBefore := tmpEnvDiff.state.GetBalance(coinbase)
+		gasUsed uint64
 
-	profitBefore := new(big.Int).Set(tmpEnvDiff.newProfit)
-	var gasUsed uint64
+		hasCommitErr     bool
+		hasReceiptFailed bool
```
Ruteri, line 261: Is `hasReceiptFailed` used in this scope? If not, it should not be defined here and instead in the scope it's actually necessary in. Same for `hasCommitErr` and `isRevertibleTx`

miner/algo_common.go
```diff
@@ -281,15 +296,28 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
+
 		receipt, _, err := tmpEnvDiff.commitTx(tx, chData)
 
-		if err != nil {
-			log.Trace("Bundle tx error", "bundle", bundle.OriginalBundle.Hash, "tx", tx.Hash(), "err", err)
+		hasCommitErr = err != nil
+		hasReceiptFailed = receipt != nil && receipt.Status == types.ReceiptStatusFailed
+
+		// if drop enabled, and revertible tx has error on commit, we skip the transaction and continue with next one
+		if algoConf.DropRevertibleTxOnErr && isRevertibleTx && hasCommitErr {
```
Ruteri, line 308: I think it'd be more readable to exhaust all possibilities for err/receipt/DropRevertibleTxOnErr/RevertingHash in a big switch or if-else. There's like four cases so it's not trivial any more

miner/algo_common.go
```diff
@@ -301,7 +329,16 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
 
 	bundleProfit := coinbaseBalanceDelta
 
-	bundleActualEffGP := bundleProfit.Div(bundleProfit, big.NewInt(int64(gasUsed)))
+	var (
+		gasUsedBigInt     = new(big.Int).SetUint64(gasUsed)
+		bundleActualEffGP *big.Int
+	)
+	if gasUsed == 0 {
```
Ruteri, line 336: Can this ever be true? I find this a bit weird. Maybe you should use an assert instead if this should never be the case but you don't want to do the division without checking.

miner/algo_common.go
```diff
@@ -301,7 +329,16 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
 
 	bundleProfit := coinbaseBalanceDelta
 
-	bundleActualEffGP := bundleProfit.Div(bundleProfit, big.NewInt(int64(gasUsed)))
+	var (
```
Ruteri, line 332: I find `gasUsedBigInt := new(big.Int).SetUint64(gasUsed)` clearer than using var blocks, and it's the convention in the codebase

miner/algo_common.go
```diff
@@ -182,17 +188,18 @@ func applyTransactionWithBlacklist(signer types.Signer, config *params.ChainConf
 
 // commit tx to envDiff
 func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData) (*types.Receipt, int, error) {
-	header := envDiff.header
-	coinbase := &envDiff.baseEnvironment.coinbase
-	signer := envDiff.baseEnvironment.signer
+	var (
```
Ruteri, line 191: Why tho

miner/algo_common.go
```diff
@@ -519,7 +556,9 @@ func (envDiff *environmentDiff) commitSBundle(b *types.SimSBundle, chData chainD
 	return nil
 }
 
-func (envDiff *environmentDiff) commitSBundleInner(b *types.SBundle, chData chainData, interrupt *int32, key *ecdsa.PrivateKey) error {
+func (envDiff *environmentDiff) commitSBundleInner(
```
Ruteri, line 559: Again, convention of the codebase is otherwise and without both a discussion and a good reason we shouldn't change it

miner/algo_common.go
Wazzymandias: yes, unit test was failing without this line

miner/algo_common.go
Wazzymandias: it's standard go practice to group multiple definitions into var blocks

miner/algo_common.go
Wazzymandias: it's interesting to define them in the scope of a for loop, but the compiler will take care of it so i can update

in this case i'll do it but there are cases where what you're describing would be less than ideal

miner/algo_common.go
Wazzymandias: the codebase is large, there are exceptions to the claim that var blocks are not the convention

see: execution.go, transition.go, runner.go, state_processor.go

miner/algo_common.go
Wazzymandias: updated

miner/algo_common.go
Wazzymandias: updated

miner/algo_common.go
Wazzymandias: updated

miner/algo_common.go
Wazzymandias: it's not failing anymore, i think because logic changed, updated to no longer check for 0 since tests are passing

miner/algo_common.go
Wazzymandias: the var block doesn't make sense anymore since we no longer check for zero, moved back to existing convention

miner/algo_greedy.go
```diff
@@ -20,17 +20,22 @@ type greedyBuilder struct {
...
 	interrupt        *int32
+	algoConf         algorithmConfig
 }
 
 func newGreedyBuilder(
-	chain *core.BlockChain, chainConfig *params.ChainConfig,
+	chain *core.BlockChain, chainConfig *params.ChainConfig, algoConf *algorithmConfig,
 	blacklist map[common.Address]struct{}, env *environment, key *ecdsa.PrivateKey, interrupt *int32,
 ) *greedyBuilder {
+	if algoConf == nil {
```
Ruteri, line 30: I'd suggest requiring algoConf to be non-nil, and instead passing `&defaultAlgorithmConfig` wherever default should be used

miner/algo_common.go
Wazzymandias: updated

miner/algo_common.go
```diff
@@ -240,15 +247,23 @@ func (envDiff *environmentDiff) commitTx(tx *types.Transaction, chData chainData
...
-	profitBefore := new(big.Int).Set(tmpEnvDiff.newProfit)
-	var gasUsed uint64
+		gasUsed uint64
+	)
 
 	for _, tx := range bundle.OriginalBundle.Txs {
+		var (
+			txHash = tx.Hash()
+			// update isRevertibleTx if transaction is found in list of reverting transaction hashes
+			isRevertibleTx = bundle.OriginalBundle.RevertingHash(txHash)
```
Ruteri, line 265: This is actually expensive, and not needed most of the time - I wouldn't compute this unless necessary

Ruteri (approved):
Deferring decision on the conventions to @dvush and @avalonche since they frequent this codebase more than I do

miner/algo_common.go
Wazzymandias: good catch, updated based on offline discussion

miner/worker.go
```diff
@@ -1944,7 +1949,10 @@ func containsHash(arr []common.Hash, match common.Hash) bool {
 
 // Compute the adjusted gas price for a whole bundle
 // Done by calculating all gas spent, adding transfers to the coinbase, and then dividing by gas used
-func (w *worker) computeBundleGas(env *environment, bundle types.MevBundle, state *state.StateDB, gasPool *core.GasPool, pendingTxs map[common.Address]types.Transactions, currentTxCount int) (simulatedBundle, error) {
+func (w *worker) computeBundleGas(
```
avalonche, line 1952: can consider restructuring these args into a struct

cmd/utils/flags.go
```diff
@@ -845,6 +845,15 @@ var (
 		Category: flags.BuilderCategory,
 	}
 
+	BuilderDiscardRevertibleTxOnErr = &cli.BoolFlag{
+		Name: "builder.discard_revertible_tx_on_error",
```
avalonche, line 849: could you update the docs in the README?

miner/algo_common.go
```diff
@@ -56,13 +55,16 @@ func (e *lowProfitError) Error() string {
...
+	DropRevertibleTxOnErr bool
+
 	// EnforceProfit is true if we want to enforce a minimum profit threshold
 	// for committing a transaction based on ProfitThresholdPercent
 	EnforceProfit bool
-	// ExpectedProfit should be set on a per transaction basis when profit is enforced
-	ExpectedProfit *big.Int
 	// ProfitThresholdPercent is the minimum profit threshold for committing a transaction
-	ProfitThresholdPercent *big.Int
+	ProfitThresholdPercent int
```
avalonche, line 67: is value 1 = 1%? can have small comment around this

miner/algo_common.go
```diff
@@ -284,13 +295,23 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
...
+					"tx", txHash, "err", err)
+				continue
+			}
 
-		if receipt.Status != types.ReceiptStatusSuccessful && !bundle.OriginalBundle.RevertingHash(tx.Hash()) {
-			log.Trace("Bundle tx failed", "bundle", bundle.OriginalBundle.Hash, "tx", tx.Hash(), "err", err)
-			return errors.New("bundle tx revert")
+			log.Trace("Bundle tx error", "bundle", bundle.OriginalBundle.Hash, "tx", txHash, "err", err)
+			return err
+		} else if receipt != nil && receipt.Status == types.ReceiptStatusFailed {
```
avalonche, line 308: should it return error on receipt == nil?

miner/algo_common.go
```diff
@@ -284,13 +295,23 @@ func (envDiff *environmentDiff) commitBundle(bundle *types.SimulatedBundle, chDa
 		receipt, _, err := tmpEnvDiff.commitTx(tx, chData)
 
 		if err != nil {
-			log.Trace("Bundle tx error", "bundle", bundle.OriginalBundle.Hash, "tx", tx.Hash(), "err", err)
-			return err
-		}
+			isRevertibleTx := bundle.OriginalBundle.RevertingHash(txHash)
+			// if drop enabled, and revertible tx has error on commit, we skip the transaction and continue with next one
+			if algoConf.DropRevertibleTxOnErr && isRevertibleTx {
+				log.Info("Found error on commit for revertible tx, but discard on err is enabled so skipping.",
```
avalonche, line 301: why is this log.Info? other logs is Trace

avalonche (approved):

miner/algo_common.go
Wazzymandias: @Ruteri suggested `log.Info` because this is expected behavior, previous discussion: https://github.com/flashbots/builder/pull/81#discussion_r1252676030

dvush (approved):
looks correct to me
re: styling - I think breaking lines looks better but no strong opinion this or that way

cmd/utils/flags.go
Wazzymandias: good catch, updated

miner/algo_common.go
Wazzymandias: good question - this should be caught in the `commitTx` function

The current pattern is that receipt is nil when error is returned

miner/algo_common.go
Wazzymandias: good point, updated

miner/algo_common.go
Ruteri: For the record, I proposed having it at Info rather than Warn - I'd be fine with debug/trace

miner/algo_common.go
Ruteri: could always put an assertion I guess

Wazzymandias: refactored it to return error on nil receipt

miner/algo_common.go
Wazzymandias: updated to debug level - this log shouldn't occur very often given current data (~10% requests specify canRevert) and doesn't offer verbose info like trace, so set to debug for now

Wazzymandias: will revisit if things change

miner/algo_greedy.go
Wazzymandias: didn't address in this PR but I made that change in another PR because it was relevant for some other changes

miner/algo_common.go
Wazzymandias: i deployed for testing and observed that debug log for this actually happens a lot - updated  to trace as @avalonche originally suggested

PR #83, opened 2023-07-06
[trivial] Parameterize greedy-buckets price cutoff percent & update bundle profit to coinbase delta

Wazzymandias:
## 📝 Summary

- Add support for `--miner.price_cutoff_percent`, used when `--miner.algotype=greedy-buckets` to support dynamic bucket sizes for block building algorithm
  - Example: Setting `--miner.price_cutoff_percent` to `0` and `--miner.algotype=greedy-buckets` is comparable to "total profit" block building algorithm, where all transactions are sorted by effective gas price, and then sorted by max profit. 

## 📚 References

---

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -548,7 +548,7 @@ var (
 	}
 	MinerAlgoTypeFlag = &cli.StringFlag{
 		Name:     "miner.algotype",
-		Usage:    "Block building algorithm to use [=mev-geth] (mev-geth, greedy)",
+		Usage:    "Block building algorithm to use [=mev-geth] (mev-geth, greedy, greedy-buckets)",
 		Value:    "mev-geth",
 		Category: flags.MinerCategory,
 	}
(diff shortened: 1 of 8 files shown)


cmd/utils/flags.go
```diff
@@ -591,6 +591,17 @@ var (
...
 		Category: flags.MinerCategory,
 	}
+	MinerPriceCutoffPercentFlag = &cli.IntFlag{
+		Name: "miner.price_cutoff_percent",
+		Usage: "flashbots - The minimum effective gas price threshold used for bucketing transactions by price. " +
+			"For example if the top transaction in a list has an effective gas price of 1000 wei and price_cutoff_percent " +
+			"is 10 (i.e. 10%), then the minimum effective gas price included in the same bucket as the top transaction " +
+			"is (1000 * 10%) = 100 wei.\n" +
+			"NOTE: This flag is only used when miner.algotype=greedy-buckets",
+		Value:    -1,
```
avalonche, line 601: why not have this at `defaultPriceCutoffPercent`? having default as -1 is a bit misleading when using `geth --help` as that is not actually the default set in the code

avalonche (approved):

cmd/utils/flags.go
Wazzymandias: that's a really good point, updated to use `defaultPriceCutoffPercent`

PR #84, opened 2023-07-08
Init BuilderRateLimitResubmitInterval builder config param from command line flag.

vazha:
## 📝 Summary

Because of absent initialization --builder.block_resubmit_interval param always ignored and default value is used

## 📚 References

---

* [ x] I have seen and agree to [`CONTRIBUTING.md`](https://github.com/flashbots/builder/blob/main/CONTRIBUTING.md)


Wazzymandias (approved):
Nice catch!

PR #93, opened 2023-08-07
Allow alternative methods of proposer payments in validation api (simple version).

dvush:
## 📝 Summary

This will allow validation API to accept blocks where:

* proposer payment is plain transfer to the fee recipient
* proposer payment is internal payment to the fee recipient made through the smart contract
* fee recipient is set to coinbase and accepts payment directly

Its optional and should be enabled by setting `--builder.validation_force_last_tx_payment=false`

This is similar to https://github.com/flashbots/builder/pull/78 but much simpler and does not discard withdrawals from the profit calculation. 

The reason that this approach it better is that:
* its simpler
* builders and fee recipients can adapt to these rules
* its already deployed in the wild by other relays so if there would be some downsides to this approach it should be already noticed

The way we calculate profit is:

* check balance difference of the fee recipient after the block is applied
* if claimed profit is bigger or equal to balance difference we accept it
* otherwise check if last tx is payment to the fee recipient with value of exact profit (previous behavior)

Note that:
* withdrawals from the fee recipient decrease profit of the bid
* deposits to the fee recipient address can be claimed as a profit

## 📚 References

---

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2494,7 +2494,11 @@ func (bc *BlockChain) SetBlockValidatorAndProcessorForTesting(v Validator, p Pro
 	bc.processor = p
 }
 
-func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Address, expectedProfit *big.Int, registeredGasLimit uint64, vmConfig vm.Config) error {
+// ValidatePayload validates the payload of the block.
+// It returns nil if the payload is valid, otherwise it returns an error.
+//   - `forceLastTxPayment` if set to true, proposer payment is assumed to be in the last transaction of the block
+//     otherwise we use proposer balance changes after the block to calculate proposer payment (see details in the code)
+func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Address, expectedProfit *big.Int, registeredGasLimit uint64, vmConfig vm.Config, forceLastTxPayment bool) error {
 	header := block.Header()
 	if err := bc.engine.VerifyHeader(bc, header, true); err != nil {
 		return err
@@ -2527,10 +2531,15 @@ func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Ad
 	// and dangling prefetcher, without defering each and holding on live refs.
 	defer statedb.StopPrefetcher()
 
+	feeRecipientBalanceBefore := statedb.GetBalance(feeRecipient)
+
 	receipts, _, usedGas, err := bc.processor.Process(block, statedb, vmConfig)
 	if err != nil {
 		return err
 	}
+
+	feeRecipientBalanceDelta := new(big.Int).Set(statedb.GetBalance(feeRecipient))
+	feeRecipientBalanceDelta.Sub(feeRecipientBalanceDelta, feeRecipientBalanceBefore)
 
 	if bc.Config().IsShanghai(header.Time) {
 		if header.WithdrawalsHash == nil {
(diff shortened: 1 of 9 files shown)


core/blockchain.go
```diff
@@ -2527,11 +2531,16 @@ func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Ad
 	// and dangling prefetcher, without defering each and holding on live refs.
 	defer statedb.StopPrefetcher()
 
+	feeRecipientBalanceBefore := statedb.GetBalance(feeRecipient)
```
Ruteri, line 2534: Underneath this function does the following:
```
func (s *stateObject) Balance() *big.Int {
	return s.data.Balance
}
```

and none of the data is copied, so here you hold the actual balance pointer of the object. I'd propose to copy it out of precaution so that people using the variable don't mistakenly change the actual balance in statedb, like you do with feeRecipientBalanceDelta. I think the optimization here is too risky

core/blockchain.go
```diff
@@ -2554,6 +2563,18 @@ func (bc *BlockChain) ValidatePayload(block *types.Block, feeRecipient common.Ad
 		return err
 	}
 
+	// Validate proposer payment
+
+	if !forceLastTxPayment {
```
Ruteri, line 2568: Hm, do we always expect the coinbase to be set to the proposer's fee recipient in this case? Maybe it'd be worth adding the check to make sure people aren't doing something more convoluted like internal transfers without consideration.
I'd be open to removing that check if people want us to, but I would be on the safer side if possible (either last tx payment or coinbase set to fee recipient)

Ruteri (approved):

core/blockchain.go
dvush: sure, why not

btw we use the same pattern for bundle profit in our builder

today it works if you don't change this variable because in geth they chose to return pointer here but when balance is changed new value is created

but who knows what change can happen later

core/blockchain.go
dvush: We also want to specifically handle case where payment is done in the last tx but it goes through some contract that can check block number and batch mev share kickbacks. 

Code that we have here handles both this case and coinbase case.

If we want to add restriction like "if coinbase is not the fee recipient than you have to pay with last tx but it also can be in internal transfer" this will make this code harder because you need to trace indivdual txs again (previous version of this pr) and it can be even more dangerous because of potential bugs and more complex logic.

dvush:
@Wazzymandias , you are probably right there could be edge-cases with reorgs. I don't think that they are critical but definitely something to remember.

dvush:
Here are some clarifications on why we don't worry about withdrawals and deposits to the fee recipient directly (as opposed to payments made to coinbase) influencing the value of the bid.

Firstly, why might this be a problem? The fee recipient can send transactions out and can also receive ETH from transactions as part of some on-chain activity unrelated to its role as the fee recipient in some blocks. According to this rule for value calculation, these transactions will influence bid value: transactions from the fee recipient, when included, will reduce bid value, and transactions to the fee recipient will increase it.

Trying to account for this leads to complicated solutions for validation and block building. By giving up on these issues, we make the implementation safer and simpler.

Below, we consider the worst-case scenario of what can happen with these changes.

The worst-case scenario is that some builders who are unaware of this rule will be at a disadvantage compared to builders who know about it. The proposer will potentially miss out on profits compared to the value of deposits/withdrawals of the transaction by that fee recipient. Also, there's an edge case of a "forwarder fee recipient" (a contract that sends all received ETH elsewhere) which might not exist in practice.

There are multiple factors that need to align for this to be a problem:

* "Some builders who don't know about this rule": If there are at least a few builders who are aware of this, they will compete with each other to pay the highest amount to the fee recipient, even if they outcompete the builders who are unaware.
* "Will potentially miss profits compared to the value of deposits/withdrawal of the transaction": The value of the deposit/withdrawal should be significant for the loss to be substantial.
* "Of the transaction by that fee recipient": There should be a transaction by the fee recipient in the mempool when this fee recipient proposes a block. For small stakers, the probability of this happening is minuscule. The probability to win a block is 1/(number of proposers), and there must be a transaction in this specific block. This only matters for staking pools, but they already use specialized addresses to collect fees .

Other points:
* Proposer pools can make frequent withdrawals from the fee recipient, and these withdrawals reduce block profit. Withdrawal transactions decrease the block profit, so builders will not include them for blocks with the same fee recipient. Unless this pool control all the blocks, withdrawal transactions will be included by others proposers who have a different fee recipient.
* "Forwarder fee recipient": It's unclear if this case exists in practice. But if there is a builder who pays to the fee recipient in the last transaction, they will cover this case because we support a fallback to the old behavior.

PR #96, opened 2023-08-10
Check bundle atomicity

dvush:
## 📝 Summary

Sanity check that all bundles that were included in the block are included preserving promised invariant.

---

### DIFF
--- a/miner/verify_bundles.go
+++ b/miner/verify_bundles.go
@@ -1,0 +1,211 @@
+package miner
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// ErrBundleTxNotFound is returned when a tx is not found in the resulting block
+type ErrBundleTxNotFound struct {
+	BundleHash common.Hash
+	TxHash     common.Hash
+	// Index of the tx in the bundle
+	TxIndex int
+}
+
+func NewErrBundleTxNotFound(bundleHash, txHash common.Hash, txIndex int) *ErrBundleTxNotFound {
+	return &ErrBundleTxNotFound{
+		BundleHash: bundleHash,
+		TxHash:     txHash,
+		TxIndex:    txIndex,
+	}
+}
+
+func (e *ErrBundleTxNotFound) Error() string {
+	return fmt.Sprintf("tx from included bundle not found tx_hash=%s, bundle_hash=%s, tx_bundle_index=%d", e.TxHash.Hex(), e.BundleHash.Hex(), e.TxIndex)
+}
+
+// ErrBundleTxReverted is returned when a tx is reverted in the resulting block, but it was not allowed to be reverted
+type ErrBundleTxReverted struct {
+	BundleHash common.Hash
+	TxHash     common.Hash
+	// Index of the tx in the bundle
+	TxIndex int
+}
+
+func NewErrBundleTxReverted(bundleHash, txHash common.Hash, txIndex int) *ErrBundleTxReverted {
+	return &ErrBundleTxReverted{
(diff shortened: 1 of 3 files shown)


miner/verify_bundles.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		if _, ok := includedTxDataByHash[hash]; ok {
+			return NewErrPrivateTxFromFailedBundle(priv.bundleHash, hash, priv.index)
+		}
+	}
+
+	return nil
+}
+
+func extractBundleTxDataFromBundles(bundles []types.SimulatedBundle, result map[common.Hash][]bundleTxData) {
```
Ruteri, line 212: Minor suggestion: I'd expect result as an outparam to be a pointer to the map because I never know if the map is a reference under the copied value

miner/verify_bundles.go
```diff
@@ -0,0 +1,285 @@
...
+				if tx.canRevert {
+					continue
+				} else {
+					return NewErrBundleTxNotFound(bundleHash, tx.hash, bundleIdx)
+				}
+			}
+
+			// we allow gaps between txs in the bundle,
+			// but txs must be in the right order
+			if txInclusion.index < currentBlockTx {
```
Ruteri, line 188: Nice

miner/worker.go
```diff
@@ -1533,8 +1546,13 @@ func (w *worker) generateWork(params *generateParams) (*types.Block, *big.Int, e
...
 	orderCloseTime := time.Now()
 
-	blockBundles, allBundles, usedSbundles, err := w.fillTransactionsSelectAlgo(nil, work)
+	blockBundles, allBundles, usedSbundles, mempoolTxHashes, err := w.fillTransactionsSelectAlgo(nil, work)
+
+	if err != nil {
+		return nil, nil, err
+	}
 
+	err = VerifyBundlesAtomicity(work, blockBundles, allBundles, usedSbundles, mempoolTxHashes)
```
Ruteri, line 1555: Should we consider putting this behind a flag maybe? I think we always want to run with this check, but who knows

miner/verify_bundles.go
```diff
@@ -0,0 +1,285 @@
...
+type privateTxData struct {
+	bundleHash common.Hash
+	index      int
+}
+
+// checkBundlesAtomicity checks that all txs from the included bundles are included in the block correctly
+// 1. We check that all non-reverted txs from the bundle are included in the block and are not reverted
+// 2. Reverted txs are allowed to be not included in the block
+// 3. All txs from the bundle must be in the right order, gaps between txs are allowed
+func checkBundlesAtomicity(
```
Ruteri, line 134: Would you maybe want to cross-check with the block itself?

miner/verify_bundles.go
dvush: I agree but pointer to map does not support `m[key]` operation so you would still need to do `(*m)[key]` below.

miner/verify_bundles.go
dvush: I use `environment` from `worker.go` to build input args to this function. Since `environment` is used to build block it should be as close to the block as possible. (we can't use `Block` because it does not have receipts inside)

miner/worker.go
dvush: int non parallel

and in your case its still 1 ms

dvush: btw you don't need to add 1000 committed bundles.  even 100 is way to big for committed stuff.

but for used bundles its more correct

miner/verify_bundles.go
Ruteri: I think that's fine

Ruteri (approved):

Wazzymandias (approved):

PR #102, opened 2023-08-21
Efficient reverts clean dirty state cache

dvush:
## 📝 Summary

Fixes bug in efficient reverts branch.

There is a bad interaction between our multi tx snapshot and journal that is used to revert changes for internal calls. That was happening because accounts that were changed and then reverted according to journal were not considered "dirty" when finalizing so their dirty state was not cleaned.

Later this state would be used for state root calculation when finalized for each object is called too. 

This adds a unit test that was failing before these changes were introduced.

---


Wazzymandias (approved):
🔥

PR #112, opened 2023-09-05
Bundle new UUID

TymKh:
## 📝 Summary

Groundwork for using unique deterministic ID for bundles

Example of new inserted bundle from goerli (json extracted from db, stripped from irrelevant data):
```
[
  {
    "id": 47188143,
    "bundle_hash": "0xc7c4bfddd2210bf883e5969c0d7824333f3b52433a452362c2b7b0dd31921f42",
    "param_block_number": 9732877,
    "error_message": null,
    "gas_fees": 0,
    "has_error": false,
    "version": 2,
    "param_reverting_tx_hashes": "",
    "is_high_prio": true,
    "owner_queue": "6cqwv",
    "replacement_uuid": null,
    "bundle_uuid": "000009a2-b7b1-5de3-8e35-7e6ba2209873"
  }
] 
```
## 📚 References

---

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -774,6 +774,7 @@ type LatestUuidBundle struct {
 	Uuid           uuid.UUID
 	SigningAddress common.Address
 	BundleHash     common.Hash
+	BundleUUID     uuid.UUID
 }
 
 type MevBundle struct {
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -19,9 +19,11 @@ package types
 import (
 	"bytes"
 	"container/heap"
+	"encoding/binary"
 	"errors"
 	"io"
 	"math/big"
+	"sort"
 	"sync/atomic"
 	"time"
 
@@ -30,6 +32,7 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/google/uuid"
+	"golang.org/x/crypto/sha3"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


core/types/transaction.go
```diff
@@ -784,6 +787,23 @@ type MevBundle struct {
...
+		return bytes.Compare(b.RevertingTxHashes[i][:], b.RevertingTxHashes[j][:]) <= 0
+	})
+	for _, txHash := range b.RevertingTxHashes {
+		buf = append(buf, txHash[:]...)
+	}
+	return buf
+}
+
+func (b *MevBundle) ComputeUUID() uuid.UUID {
+	return uuid.NewHash(sha3.NewLegacyKeccak256(), uuid.Nil, b.UniquePayload(), 5)
```
Wazzymandias, line 804: question: Why are you using legacy keccak here?

TymKh: I was thinking among the lines of unity with existing geth codebase/possible js tooling. If not for this I would just go for sha256.

Wazzymandias: yes, please use sha256 or blake2b, I don't think we need legacy keccak here - that's intended for very specific case

core/types/transaction.go
```diff
@@ -784,6 +788,23 @@ type MevBundle struct {
...
+	sort.Slice(b.RevertingTxHashes, func(i, j int) bool {
+		return bytes.Compare(b.RevertingTxHashes[i][:], b.RevertingTxHashes[j][:]) <= 0
+	})
+	for _, txHash := range b.RevertingTxHashes {
+		buf = append(buf, txHash[:]...)
+	}
+	return buf
+}
+
+func (b *MevBundle) ComputeUUID() uuid.UUID {
```
Wazzymandias, line 804: suggestion (non-blocking)

We should probably event use a `sync.Once` pattern to initialize the UUID once since we don't mutate bundles after initialization

Wazzymandias:
looks great! 

Could you update your [PR description](https://github.com/flashbots/builder/pull/112#issue-1881589865) with output successfully showing bundles inserted into database with the new schema and UUID? 

Otherwise lgtm, good stuff

core/types/transaction.go
TymKh: Yeah, I thought in a little bit different direction, I believe we'll eventually move to propagating this uuid as a field of bundle structure, but it'd be better be done once we have some module with shared models. But I anyway understand your concern, I hope we'll address it in future

TymKh:
> looks great!
> 
> Could you update your [PR description](https://github.com/flashbots/builder/pull/112#issue-1881589865) with output successfully showing bundles inserted into database with the new schema and UUID?
> 
> Otherwise lgtm, good stuff

Done

Wazzymandias (approved):

PR #113, opened 2023-09-06
Feature: overwrite request gas limits and return new values to relay

prestwich:
## 📝 Summary

Modifies the v2 endpoint to overwrite gaslimit during simulation, and to return the overwritten values to the relay client

This allows the block-validation logic in the API to adjust the gaslimit set by the block builder in order to achieve specific gas targets specified by the relay api service.

## 📚 References

This is in support of https://github.com/flashbots/mev-boost-relay/pull/519

---


prestwich:
hard to tell because of the large output, but given the latest commit is non-functional changes it seems likely the CI failure is unrelated

PR #123, opened 2023-10-03
Add fix and update unit tests for header gas limit calculation 

Wazzymandias:
## 📝 Summary

- Update builder submission logic to calculate gas limit on payload attribute
- Addresses long standing issue of blocks being rejected by relay on block validation due to invalid gas limit set on header
- Tested with unit tests and on deployed service for several weeks to confirm the changes fix observed issue

## 📚 References

- [Should builder block submissions set a different gas limit than what proposer specified?](https://collective.flashbots.net/t/should-builder-block-submissions-set-a-different-gas-limit-than-what-proposer-specified/2414)
---

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/ethereum/go-ethereum/core"
 	"math/big"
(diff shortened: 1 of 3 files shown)


Ruteri (approved):
LGTM

PR #126, opened 2023-10-11
Correct estimation of timestamp in eth_callBundle

dvush:
Before we used parent block timestamp + 1 and this breaks simulation of bundles that check that timestamp of the next target block is what is expected.

## 📚 References

---


Wazzymandias (approved):
😵‍💫

PR #141, opened 2024-01-30
return proper error on callBundle incorrect gasPrice

TymKh:
## 📚 References

---


internal/ethapi/api.go
```diff
@@ -2385,7 +2385,7 @@ func (s *BundleAPI) CallBundle(ctx context.Context, args CallBundleArgs) (map[st
 		}
 		uint256GasPrice, ok := uint256.FromBig(gasPrice)
 		if !ok {
-			return nil, fmt.Errorf("err: %w; txhash %s", err, tx.Hash())
```
dvush, line 2388: I feel like someone can depend on this particular message struct. 

Is it possible to not change it?

0x416e746f6e (approved):

internal/ethapi/api.go
0x416e746f6e: what if:

```go
return nil, fmt.Errorf("err: incorrect gasPrice; txhash %s", tx.Hash())
```

dvush: ok, as I undestood this is the code path that was added for deneb so no need to make this compatible

TymKh:
problem was actually deeper than this, I pushed another fix

PR #144, opened 2024-02-19
Exclude CL withdrawals from profit calculation

dvush:
## 📝 Summary

This adds a new flag `--builder.validation_exclude_withdrawals` that changes how validation API calculates profit of the bid.

If enabled block validation API will exclude CL withdrawals to the fee recipient from the balance delta that is used to calculate profit of the fee recipient.

---


avalonche (approved):

avalonche (commented):
I think it should use the `WITHDRAWALS_SCORING_ACTIVATION_EPOCH` as a flag instead to determine when to exclude withdrawals

Ruteri (approved):

dvush:
* relays decided to enable it when they want not on the epoch but approximately before dencun
* linter is horribly broken and needs closer investigation anyway

PR #162, opened 2024-06-26
Validation jun 10

Ruteri:
## 📝 Summary

Fixes an unsigned integer underflow bug in coinbase balance diff calculation.

## 📚 References

https://collective.flashbots.net/t/disclosure-underflow-bug-in-mev-boost-relay/3599

---


metachris (approved):

eth/block-validation/api_test.go
```diff
@@ -845,6 +845,99 @@ func executableDataToBlockValidationRequest(execData *engine.ExecutableData, pro
 	return blockRequest, nil
 }
 
+func TestValidateBuilderSubmissionV2_CoinbasePaymentUnderflow(t *testing.T) {
```
avalonche, line 848: should use the v3 builder submission which is currently being used, though ideally the unit test should just be calling `ValidatePayload`

avalonche (approved):

